PR #2, opened 2022-12-01
Sys 1095/create item status model

kjallen:
This PR is for initial add of the `ItemStatus` model, taking the place of the `QA_STATUS` table in DLCS. 
The size of fields was chosen based on the content in the existing table and previous definition.
A fixture has been added for initial population of data, and only needs to be run once on initial data load.

To verify content:

Run migrations:
`$ docker-compose exec django python manage.py migrate`

Load table data:
`$ docker-compose exec django python manage.py loaddata seed-data.json`

Retrieve data in REPL:
```
>>> from oh_staff_ui.models import ItemStatus
>>> all = ItemStatus.objects.all()
>>> all.values()

<QuerySet [{'id': 1, 'status': 'In progress', 'status_description': 'the record is being actively catalog.'}, {'id': 2, 'status': 'Completed', 'status_description': 'the resource and the metadata record are complete and ready to be part of the collection.'} 
...
```

There should be 12 records total, copied from the content in `QA_STATUS` table. 
The initial data matches content as exists and may be corrected in future to reflect current use, future imports will overwrite existing entries using the same primary key.


akohler (changes requested):
@kjallen This is all good - no changes to what's here, all works as expected.  Please enable the Django admin interface and register the model so we can access / update data via the admin GUI.  Thanks --Andy

akohler (approved):
@kjallen Thanks for adding the admin stuff.  All looks good to me now, merging.

PR #3, opened 2022-12-05
Initial add of ItemType model and fixture data

kjallen:
Addition of the ItemType model, this model defines the hierarchy of the object type (note: different from parent hierarchy of items). Example hierarchy:

- Collection
    - Interview Series
        - Audio
        - Video

Nothing too exciting in this table, but please review the self-referencing key settings for agreement.
Blank or null values are permitted, the most obvious use is the top level item, which has no parent. 
A cascade delete has been set, this table, like ItemStatus, should not change much, if at all once deployed to production.

`parent = models.ForeignKey("self", on_delete=models.CASCADE, blank=True, null=True)`

To confirm properly working, please apply the following:

- Run migrate
- Run migrate loaddata
- Open admin panel to verify contents


kjallen:
Thanks for pointing out seed-data.json discrepancy, Video was corrected to point to the correct parent. 

The correct hierarchy is:

- Collection
    - Series
        - Interview
            - Audio
            - Video 

After adding the ``__str__`` overload, the ItemType.type correctly shows in the admin display

akohler (approved):

akohler:
With the data and model changes, all looks great - merging.

PR #6, opened 2023-02-07
Added Name and NameType models and related fixtures

kjallen:
Addition of Name and NameType models and partial (Name) and full (NameType) fixtures for each table.

To confirm addition and data:

- Restart application or run django migration process
    -  docker-compose exec django python manage.py migrate
- Load related data in this order:
    -  docker-compose exec django python manage.py loaddata name-type-data
    - docker-compose exec django python manage.py loaddata name-data

Navigate to the [django admin interface](http://localhost:8000/admin/) and confirm models exist and data is visible. There should be 6 NameType records and 2 Name records

A separate ticket will be created to migrate/load the entire Name list from DLCS, as it requires some additional export work to create the fixture file. 

As with the Language related tables and next dozen or so metadata tables, these are orphaned and a MtoM relationship to ProjectItems will be added later once all static data is loaded


akohler (approved):
@kjallen All seems fine, no changes needed - merging.

PR #9, opened 2023-02-22
Initial model structure for publisher and description

kjallen:
Addition of Publisher and Description models and partial fixtures for each table.

To confirm addition and data:

`docker-compose exec django python manage.py migrate`
Load related data in this order:

```
docker-compose exec django python manage.py loaddata description-type-data
docker-compose exec django python manage.py loaddata publisher-type-data
docker-compose exec django python manage.py loaddata description-data
docker-compose exec django python manage.py loaddata publisher-data
```

Navigate to the [django admin interface](http://localhost:8000/admin/) and confirm models exist and data is visible.

A separate ticket will be created to migrate/load the entire Description list from DLCS, as it requires some additional export or migration work to create the fixture file. Publisher only contains the UCLA field so will not require additional values.


akohler (approved):
@kjallen All looks good, merging

PR #10, opened 2023-03-01
SYS-1163: Management command to load ProjectItems into Django

ztucker4:
Implements [SYS-1163](https://jira.library.ucla.edu/browse/SYS-1163)

This PR creates a new Django management command, `import_projectitems`. The command takes a single argument: the name of the CSV file to input, assumed to be in the "fixtures" directory (and included in this PR). Example usage:

`docker-compose exec django python manage.py import_projectitems oh-projectitems-export-3.csv`

The current implementation loops through the set to be imported several times, so it could be made more efficient if needed. It takes <2 minutes to run for me.

Any existing ProjectItems will be deleted when the command is run. Any users (identified by email) not present in the database will be created with Django "unusable passwords". I have one specific question about my implementation in a comment.


oh_staff_ui/management/commands/import_projectitems.py
```diff
@@ -0,0 +1,101 @@
...
+            status=ItemStatus.objects.get(status=pi_dict["STATUS"]),
+            title=pi_dict["TITLE"],
+            type=ItemType.objects.get(type=pi_dict["TYPE"]),
+        )
+        p.save()
+
+    def format_date(self, date: str) -> str:
+        return datetime.datetime.strptime(date, "%d-%b-%y").strftime("%Y-%m-%d")
+
+    def format_sequence(self, seq) -> int:
```
ztucker4, line 90: Is there a better approach here? The model setup requires an integer sequence number for every ProjectItem, but the exported data has many rows with blanks here. Should empty string map to 0 or 1, or should we update the model definition to make this field optional?

akohler: Decision: Use 0 as value when blank; leave this as a required field.

oh_staff_ui/management/commands/import_projectitems.py
```diff
@@ -0,0 +1,101 @@
...
+            create_date=self.format_date(root_item["CREATE_DATE"]),
+            created_by=self.get_or_create_user(root_item["CREATED_BY"]),
+            last_modified_date=self.format_date(root_item["LAST_MODIFIED_DATE"]),
+            last_modified_by=self.get_or_create_user(root_item["LAST_MODIFIED_BY"]),
+            sequence=self.format_sequence(root_item["SEQUENCE"]),
+            status=ItemStatus.objects.get(status=root_item["STATUS"]),
+            title=root_item["TITLE"],
+            type=ItemType.objects.get(type=root_item["TYPE"]),
+        )
+        p.save()
```
akohler, line 46: @ztucker4 I don't understand why this root item gets saved to the database.  I thought we did not want to migrate this single top-level collection record?  But I see it in the database after this runs:
```
>>> p = ProjectItem.objects.get(ark="21198/zz0008zh0f")
>>> p
<ProjectItem: Oral History Collection>
>>> vars(p)
{'_state': <django.db.models.base.ModelState object at 0x7f4f34d64290>, 'id': 20454, 'ark': '21198/zz0008zh0f', 'create_date': datetime.date(2008, 1, 31), 'created_by_id': 2, 'last_modified_date': datetime.date(2009, 6, 1), 'last_modified_by_id': 3, 'parent_id': None, 'sequence': 1, 'status_id': 2, 'title': 'Oral History Collection', 'type_id': 1}
>>>
```

oh_staff_ui/management/commands/import_projectitems.py
```diff
@@ -0,0 +1,101 @@
...
+        print(f"Found {total_rows} ProjectItems to import.")
+
+        # start with known top-level ProjectItem
+        # handled outside of main loop due to lack of parent
+        added_items = []
+        root_item = next(
+            item for item in projectitem_dicts if item["ARK"] == "21198/zz0008zh0f"
+        )
+        # create and save ProjectItem for root object
+        p = ProjectItem(
```
akohler, line 35: Per my previous comment, this may not be necessary.  If it is still needed, I think it (and the `save()` at line 46) could be replaced with a call to `add_projectitem(root_item)`, though you may need to add a (blank/null) `parent` key first.

akohler (changes requested):
@ztucker4 This works fine for me, running in under a minute on my machine :)
Only concern is the creation of the top-level collection root item, which I thought we did not want.  Depending on that decision, code could be removed, or redundant code replaced.

Thanks --Andy

ztucker4:
Hi @akohler, I've updated the PR to address your comments. I reviewed your and Kristian's comments on SYS-1162, so the command now skips adding the topmost item and imports the Series items below it without a parent.

akohler (approved):

akohler:
@ztucker4 Thanks for the changes.  Confirmed the top-level collection record is no longer created, which resolves all my concerns - merging.

PR #12, opened 2023-03-03
SYS-1178: Basic item search interface

ztucker4:
Implements [SYS-1178](https://jira.library.ucla.edu/browse/SYS-1178)

Creates basic search functionality for ProjectItems. After login, the search form (http://127.0.0.1:8000/item_search/) allows users to search by ARK or Title, finding case-insensitive substring matches for each. The results are displayed on a new search_results page, sorted alphabetically by ARK or Title depending on the type of search performed. Each result ARK contains a link to the corresponding /item/item_id page.

Let me know if we should have more/different functionality at this stage (different types of result sort or search matching? more Item metadata in the results display?) - thanks!

### DIFF
--- a/oh_staff_ui/templates/oh_staff_ui/search_results.html
+++ b/oh_staff_ui/templates/oh_staff_ui/search_results.html
@@ -2,15 +2,15 @@
 
 {% block content %}
 {% if results %}
-<table style="width: 100%">
+<table class="search-results">
     <tr>
         <th>ARK</th>
         <th>Title</th>
     </tr>
     {% for item in results %}
     <tr>
-        <td style="border: 1px solid"><a href="{% url 'edit_item' item.id %}">{{ item.ark }}</a></td>
-        <td style="border: 1px solid">{{ item.title}}</td>
+        <td><a href="{% url 'edit_item' item.id %}">{{ item.ark }}</a></td>
+        <td>{{ item.title}}</td>
     </tr>
     {% endfor %}
 </table>
--- a/oh_staff_ui/templates/oh_staff_ui/search_results.html
+++ b/oh_staff_ui/templates/oh_staff_ui/search_results.html
@@ -1,0 +1,21 @@
+{% extends 'oh_staff_ui/base.html' %}
+
+{% block content %}
+{% if results %}
+<table style="width: 100%">
+    <tr>
+        <th>ARK</th>
+        <th>Title</th>
+    </tr>
+    {% for item in results %}
+    <tr>
+        <td style="border: 1px solid"><a href="{% url 'edit_item' item.id %}">{{ item.ark }}</a></td>
+        <td style="border: 1px solid">{{ item.title}}</td>
+    </tr>
+    {% endfor %}
+</table>
+{% else %}
+<p>No results found.</p>
+{% endif %}
+
+{% endblock %}
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -12,3 +12,11 @@ class ProjectItemForm(forms.Form):
     sequence = forms.CharField(
         required=True, max_length=3, widget=forms.TextInput(attrs={"size": 3})
     )
+
+
+class ItemSearchForm(forms.Form):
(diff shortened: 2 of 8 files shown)


oh_staff_ui/templates/oh_staff_ui/search_results.html
```diff
@@ -0,0 +1,21 @@
+{% extends 'oh_staff_ui/base.html' %}
+
+{% block content %}
+{% if results %}
+<table style="width: 100%">
```
akohler, line 5: Let's avoid the use of inline CSS.  Instead, move styles (here & lines 12-13) to a new file `static/css/style.css`, similar to https://github.com/UCLALibrary/dlcs-staff-ui/blob/main/static/css/style.css , and uncomment line 6 in `oh_staff_ui/templates/oh_staff_ui/base.html`.

akohler (changes requested):
@ztucker4 This is an excellent start!  Base functionality is all there, and I like the way you provided query-specific paths to search results so they can be bookmarked / shared.

One minor change to CSS handling, and one larger change to how title searches are done, and then this can move forward.

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for re-review. CSS has been moved into a separate file and keyword search is implemented for title. This is done by splitting the search query at each space character and ORing the substring filter results for each word. Testing example:
* [search for "bart"](http://127.0.0.1:8000/search_results/title/bart): 2 results
* [search for "lisa"](http://127.0.0.1:8000/search_results/title/lisa): 9 results
* [search for "maggie"](http://127.0.0.1:8000/search_results/title/maggie): 1 result
* [search for "bart lisa maggie"](http://127.0.0.1:8000/search_results/title/bart%20lisa%20maggie): 12 results, superset of previous results

I realized while writing this there may be other/better ways to implement "keyword" search, like by limiting to whole word matches or using the fancy [Django-postgresql search module](https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/search/) - but maybe this is a possible future enhancement.

akohler:
@ztucker4 , I'm not sure ORing each term is ideal.  (Some search engines do that... not sure they're right either :) ).  With your examples, I would expect a search for "bart lisa maggie" to return no records, instead of all of them.

For another example, consider http://127.0.0.1:8000/search_results/title/session%2013

With the implemented `OR`, I get hundreds of results, a combination of the `OR` and the substring search for `13`.  Since dealing with whole words is harder, let's defer that as an enhancement for later. Switching to '&` for `AND` finds... nothing.  Hmmm, that may be due to the inclusion of the original query string along with the individual keywords.

Maybe a hybrid approach, which searches for (a) the original full substring OR (b) the AND-ed set of keywords?  I'm testing this locally now:
```
def construct_keyword_query(query: str) -> Q:
    # Always include the full query, as a single substring.
    full_q = Q(title__icontains=query)
    keywords = query.split()
    # If there's only one word, it's the same as the full query, nothing to do.
    # Otherwise, grab the first word, then AND each following word.
    if len(keywords) > 1:
        words_q = Q(title__icontains=keywords[0])
        # All except the first word
        for word in keywords[1:]:
            words_q = words_q & Q(title__icontains=word)
        # OR the assembled set of words with the full query.
        full_q = full_q | words_q
    return full_q
```
Using http://127.0.0.1:8000/search_results/title/session%2013
This hybrid finds the dozens of records which have both terms.

http://127.0.0.1:8000/search_results/title/session%13
This hybrid finds the same set as "session 2013" above, due to the substring search, along with the 3 records which have the literal string "session 13" in the titles.

Enough for now, let's talk about this Monday - happy weekend!  Thanks --Andy

ztucker4:
Hi @akohler, I've updated the PR with the new keyword search logic - (full query) OR (keywords ANDed together). Ready for re-review, thanks!

akohler (approved):

akohler:
@ztucker4 Thanks for making the changes.  There's a problem searching for full ARKs like `21198/zz002h4hmn`, but that fix can wait until later - I'll open a new bug ticket for that.  Merging this now.

PR #13, opened 2023-03-03
SYS-1182: Modify metadata models to support item associations

akohler:
Implements [SYS-1182](https://jira.library.ucla.edu/browse/SYS-1182).

@kjallen Please review this one.  @ztucker4 FYI.

This PR adds 7 new "usage" models, to associated `ProjectItem` with current optional metadata models in many-to-many relations.  Where relevant (all except `Language`), the corresponding "type" field moves as well.
* `AltId`: `ItemAltIdUsage`
* `AltTitle`: `ItemAltTitleUsage`
* `Description`: `ItemDescriptionUsage`
* `Language`: `ItemLanguageUsage`
* `Name`: `ItemNameUsage`
* `Publisher`: `ItemPublisherUsage`
* `Subject`: `ItemSubjectUsage`

`admin.py` is updated to remove the `type` field from display for the base models, and the sample data files in `fixtures` are updated to remove `type` data (e.g., the 2 sample records in `name-data.json` no longer have `type` values, etc.).

I did not add these new usage models to `admin.py`, as I don't think there's a need to see that data in the Admin interface.  The usages will be created & maintained via the main item editing form(s).

### Testing
Restart application, which will apply the new migration and load the updated fixture files, replacing existing sample data.  Check logs to be sure nothing nasty happens :)
```
# If running
$ docker-compose down

$ docker-compose up -d
$ docker-compose logs -f django
```


kjallen:
Cloned clean copy, deleted local DB and containers.

No migration errors and startup, verified content via admin interface.

PR #14, opened 2023-03-08
Addition of copyright model and related fixtures

kjallen:
Addition of Copyright models and partial fixtures for each table.

To confirm addition and data:

docker-compose exec django python manage.py migrate
Load related data:

docker-compose exec django python manage.py loaddata copyright-type-data
docker-compose exec django python manage.py loaddata copyright-data

Navigate to the [django admin interface](http://localhost:8000/admin/) and confirm models exist and data is visible.


akohler (changes requested):
@kjallen This looks OK to me, but please add the two `loaddata` commands to `entrypoint.sh`, following https://github.com/UCLALibrary/oral-history-staff-ui/blob/SYS-1175/copyright_models_and_data/docker_scripts/entrypoint.sh#L36

That ensures the sample data is consistently loaded on startup (for developers).  Once real data is available, we may switch to those lists.

Thanks --Andy

kjallen:
Yes, thank you for reminder @akohler, I had forgotten that

oh_staff_ui/models.py
```diff
@@ -258,3 +258,32 @@ class ItemPublisherUsage(models.Model):
...
+
+    def __str__(self):
+        return self.value
+
+
+class ItemCopyrightUsage(models.Model):
+    item = models.ForeignKey(
+        ProjectItem, on_delete=models.PROTECT, blank=False, null=False
+    )
+    publisher = models.ForeignKey(
```
akohler, line 284: @kjallen Didn't notice this before, sorry, but `publisher` here seems wrong.  Should this be `copyright` or `copyright_statement` or something else?

akohler (changes requested):

oh_staff_ui/models.py
kjallen: Thank you for catching @akohler, corrected and updated.

akohler (approved):

PR #17, opened 2023-03-10
Addition of Relation, Coverage and Format models

kjallen:
Addition of:

- Coverage added to `ProjectItem`
- Format (not qualified, not controlled) added with model and related fixture

`docker-compose exec django python manage.py migrate`

Load data (also in entrypoint script):
 `python manage.py loaddata format-data`


oh_staff_ui/models.py
```diff
@@ -287,3 +287,61 @@ class ItemCopyrightUsage(models.Model):
...
+class ItemFormatUsage(models.Model):
+    item = models.ForeignKey(
+        ProjectItem, on_delete=models.PROTECT, blank=False, null=False
+    )
+    format = models.ForeignKey(
+        Format, on_delete=models.PROTECT, blank=False, null=False
+    )
+
+
+class Coverage(models.Model):
```
akohler, line 334: Discussing in Slack but noting here:

Is there a controlled list of Coverage values that are used?  The legacy app has UI element "Chronological period covered", which seems to be free text and single occurrence, I don't see a way to add more though maybe it repeats after the first is added?

If single-use, maybe this should become an optional attribute of ProjectItem.  If repeatable but uncontrolled, maybe we should just add ProjectItem as a FK in the Coverage table, without an ItemCoverageUsage many-to-many table.

After discussion, since this is repeatable uncontrolled free text, we don't need the Usage table, just add item as FK to `Coverage`.

oh_staff_ui/models.py
```diff
@@ -287,3 +287,61 @@ class ItemCopyrightUsage(models.Model):
...
+    )
+    relation = models.ForeignKey(
+        Relation, on_delete=models.PROTECT, blank=False, null=False
+    )
+    type = models.ForeignKey(
+        RelationType, on_delete=models.PROTECT, blank=False, null=False
+    )
+
+
+class Format(models.Model):
```
akohler, line 318: Same general question as for `Coverage` - if this is uncontrolled, there seems to be no real benefit in a `...Usage` table, so this could be an optional single attribute on `ProjectItem` or a standard `Format` model with `ProjectItem` added as an FK.

After discussion, since this is repeatable uncontrolled free text, we don't need the Usage table, just add item as FK to `Format`.

oh_staff_ui/fixtures/relation-type-data.json
```diff
@@ -0,0 +1,72 @@
+[
+    {
+        "model": "oh_staff_ui.DescriptionType",
```
akohler, line 3: These all use the `DescriptionType` model, but file (and PR) says `RelationType`.  After loading, admin interface shows nothing in the `Relation Types`, but these 10 values in `Description Types`.

oh_staff_ui/models.py
```diff
@@ -287,3 +287,61 @@ class ItemCopyrightUsage(models.Model):
     type = models.ForeignKey(
         CopyrightType, on_delete=models.PROTECT, blank=False, null=False
     )
+
+
+class RelationType(models.Model):
```
akohler, line 292: Since `RelationType` seems to have just one real value, `relatedTo`, this probably could be eliminated if we don't expect more of these types are needed.  But see other comment about `DescriptionType` vs. `RelationType` confusion.

akohler (changes requested):
@kjallen See comments, which mirror our Slack discussion.  Uncontrolled single-use optional fields could be added to `ProjectItem`; uncontrolled repeatable fields with no supporting `Type`s should stay separate, but don't need a `Usage` table as there's no benefit to trying to reuse these uncontrolled values.

oh_staff_ui/models.py
kjallen: Will add `Coverage` field directly to `ProjectItem`, this is an uncontrolled non-repeatable field

oh_staff_ui/models.py
kjallen: Format is uncontrolled but repeatable, so will add item FK to `Format`

oh_staff_ui/models.py
kjallen: I'm going to remove `Relation` and related table altogether, there is no data in existing DLCS database and the relationships on implicit in the parent/child structure of the `ProjectItem` relationships

kjallen:
- Added optional `coverage` field to `ProjectItem`
- Added `ProjectItem` fk to `Format` model
- Removed `Relation` related models and fixtures
- Corrected `format-data` fixture and entrypoint script

oh_staff_ui/admin.py
```diff
@@ -126,3 +128,8 @@ class CopyrightType(admin.ModelAdmin):
 @admin.register(Copyright)
 class Copyright(admin.ModelAdmin):
     list_display = ("value", "source")
+
+
+@admin.register(Format)
+class Format(admin.ModelAdmin):
+    list_display = ("value",)
```
akohler, line 135: Please change this to `("item", "value",)`, so we can see a link to the `ProjectItem` each is associated with.  Since these are uncontrolled values, they really only have meaning in context of the item.

akohler (changes requested):
Much simpler, I like it.  Just one minor tweak to the admin display of `Format`, then I'll merge.  Please also update the PR description to reflect just the one relevant data load.

oh_staff_ui/admin.py
kjallen: Updated to reflect, item title display in admin display

akohler (approved):

PR #18, opened 2023-03-16
SYS-1179: Add support for item editing

akohler:
Implements [SYS-1179](https://jira.library.ucla.edu/browse/SYS-1179).

This PR adds support for editing items and specific types of optional metadata (names and subjects, for now).  It establishes a basic "framework" for supporting other optional metadata:
* Define a "usage form" in `forms.py` which supports data for a "usage model".  For example, `NameUsageForm` supports the `ItemNameUsage` model.
  * Name & Subject forms use fully-controlled types and values backed by other models.
  * Some other forms will need to use controlled types, but uncontrolled values.
* Add code to `views_utils.py` which
  * Creates and populate a formset with existing values; see `get_name_formset()` and `get_subject_formset()`.
  * Adds the formset to `get_edit_item_context()`, so it can be passed to the `edit_item.html` template.
  * Saves the submitted data; see `save_item_names()` and `save_item_subjects()`.
  * Call the `save_xyz` methods from `save_all_item_data()`
 * The main view, `edit_item()` in `views.py`, should not need changing.  It passes the request to `save_all_item_data()`, builds a fresh context dictionary via `get_edit_item_context()` and passes that to the template.
 * The new `edit_item.html` template displays all relevant data and forms for adding/editing/deleting data.
   * Update the template to display each new formset in the appropriate place, following the current layout.
   * Since formset fields are being displayed individually, not via convenience templates like `form.as_p` or `form.as_div`, be sure to add each `formset.management_form` as well.

### Known limitations
* Only Name and Subject metadata is supported for now.  Other types will be added via separate tickets.
* New names & subjects must be added to items one at a time (though you can add one name and one subject at the same time), clicking the Save button.
  * Support for adding multiple names / subjects (and other types) at once will be added later, via javascript.
  * All names / subjects currently displayed can be edited and/or deleted now, with a single form submission.
* Some item metadata - Coverage, Status, and Parent - currently is not displayed and can't be edited.  This will be addressed later, via separate ticket.
* There's no "changes saved" confirmation message; I'm not currently sure how to add that, given the way the context is built and passed.

### Testing
* Load full sets of item and name data - not required, but helpful:
```
# Load full set of item data
$ docker-compose exec django python manage.py import_projectitems oh-projectitems-export-3.csv
# Load full set of name data
$ docker-compose exec django python manage.py import_names name-md-export.csv
# Other full sets of data, like subject, are not yet available.
```
* Retrieve existing item with name(s) associated.  Example: http://127.0.0.1:8000/search_results/ark/zz001bmsqw (which for me leads to http://127.0.0.1:8000/item/1234 , which has 2 names associated).
* Change type(s) - e.g., `interviewee` to `publisher` etc. - and name(s), by selecting different values from the drop down lists.
* Check the "Delete?" box to remove name associations (the "usage" records).
* Try combinations - adding a new type+name, changing existing type/name, deleting existing.  Combinations can be submitted together, subject to the limitation noted above of only adding one name at a time.
* Try subjects also, by themselves and in combination with names.
* Edit the item title / type / sequence.
* Report errors, if any...

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ The container runs via `docker_scripts/entrypoint.sh`, which
 * Waits for the database to be completely available.  This can take 5-10 seconds, depending on your hardware.
 * Applies any pending migrations (DEV environment only).
 * Creates a generic Django superuser, if one does not already exist (DEV environment only).
+* Loads fixtures to populate lookup tables and to add a few sample records.
 * Starts the Django application server.
 
 ### Setup
@@ -77,6 +78,11 @@ The container runs via `docker_scripts/entrypoint.sh`, which
    $ docker-compose exec django python manage.py migrate
    # Populate database with seed data (once it exists...)
    $ docker-compose exec django python manage.py loaddata --app oh_staff_ui seed-data
+   #
+   # Load full set of item data
+   $ docker-compose exec django python manage.py import_projectitems oh-projectitems-export-3.csv
+   # Load full set of name data
+   $ docker-compose exec django python manage.py import_names name-md-export.csv
    ```
 7. Connect to the running application via browser
 
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -1,5 +1,5 @@
 from django import forms
-from oh_staff_ui.models import ItemType
+from oh_staff_ui.models import ItemType, Name, NameType, Subject, SubjectType
 
 
 class ProjectItemForm(forms.Form):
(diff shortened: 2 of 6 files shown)


akohler:
@kjallen Testing / validation... I learned yesterday that the empty-field validation is done by the browser itself, due to the "required" HTML attribute - that happens without a trip from browser to server.

Type validation, like non-numerics for sequence, can be handled by Django.  Not sure yet if we'll write custom validators / handlers there, or basic javascript validation for client-side, or both.

kjallen:
@akohler - Thanks for that clarification, good to know.

Reviewed code - I'm still leveling up on forms, but don't have any constructive comments to add other than the already added TODO comments. 

I didn't see `create_date` or `last_modified_date` in `add_item` function, then realized being added upon initial add by model default value.

PR #19, opened 2023-03-17
SYS-1174: Add real and mock ARK support

akohler:
Implements [SYS-1174](https://jira.library.ucla.edu/browse/SYS-1174).

This PR adds a `get_ark()` method to `views_utils.py`.  This method is called by the `add_item()` view.

If running in the `dev` environment, `get_ark()` returns a mock ARK, with prefix of `FAKE/` followed by 10 random(ish) hex characters, like `FAKE/249dd89b20`.  Otherwise, `get_ark()` tries to obtain a real ARK from our ARK minter service.  This can fail - the service may be unavailable, or forbidden due to IP address restrictions - in which case an `HTTPError` is thrown and logged:
```
CRITICAL 2023-03-16 20:11:28,881 oh_staff_ui.views_utils views_utils 403 Client Error: Forbidden for url: http://noid.library.ucla.edu/noidu_zz8+?mint+1
```
The `add_item()` view wraps the creation of a new item in a `try/except` block.  If it fails (currently, only checking for `HTTPError`), no item is created.  The user remains on the `add_item` page, with the same data they entered, so they can try again.

Failure screenshot:
oh_ark_failure
Success screenshot, with fake ARK:
oh_ark_success

Other notes:
* URL of the ARK minter service is set via environment variable, so code won't need to change if that service address changes.
* Fake ARKs are used only when running in `dev` environment, not `test` or `prod`.
* Exception handling probably could be improved.
* I should write test code for this, but didn't.

### Testing
* Changing the environment via `DJANGO_RUN_ENV` in `.docker-compose_django.env` requires restarting the application after each change.
* Quicker to change the test on line 34 of `views_utils.py`, which doesn't require application restart... just be sure to revert changes.

### DIFF
--- a/.docker-compose_django.env
+++ b/.docker-compose_django.env
@@ -24,3 +24,6 @@ DJANGO_LOG_LEVEL=INFO
 DJANGO_SUPERUSER_USERNAME=admin
 DJANGO_SUPERUSER_PASSWORD=admin
 DJANGO_SUPERUSER_EMAIL=softwaredev-systems@library.ucla.edu
+
+# URL to ARK minter (PROD)
+DJANGO_ARK_MINTER=http://noid.library.ucla.edu/noidu_zz8+?mint+1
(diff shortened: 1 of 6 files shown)


kjallen (approved):
Set env variable, restarted application, confirmed request module added and fake ark added for new items

PR #21, opened 2023-03-17
Initial export file add

kjallen:
Initial add of exported metadata from OH database.

Changes:
- Addition of `export_scripts` directory at top level of project
- Contains a tab delimited file for each Descriptive Term, named by the human readable label with spaces replaced with underscores + `.tsv`
- New lines and carriage returns were present in data and are removed before writing
- Each file has 4 columns
    - ARK - The ark of the item this row is associated with
    - TYPE - The qualifier associated with this metadata, may be empty/null (example: Language)
    - VALUE - The text value of this particular metadata field
    - SOURCE - The text value of the source of this metadata field

No duplicate records of the above 4 value combinations should be present in the resulting `.tsv` file

To replicate this export:
- Create an `api_keys.py` file in the same directory with a dictionary named `api_keys` populating:
    - USER
    - PASSWORD
    - DSN
    
This must be run from a machine that has access to the database


akohler (changes requested):
@kjallen The python script itself looks fine, other than the minor file extension change.

Three of the data files have some rows with empty values.  Are these errors / problems, or actual empty values?  Do you want to filter out via query / processing here, or should we skip these when loading the data?

`for f in *.tsv; do echo "$f"; awk '{if ($3=="") print $0}' "$f" ; done`

There are empty values in 7 Chronological... rows , 2 Titles, 108 Formats - samples:
```
Chronological_period_covered.tsv
21198/zz002j75v6
21198/zz002kd86g
21198/zz002kdcm4
21198/zz002kdkhg
21198/zz002kprvn
21198/zz002knpdh
21198/zz002knt8c

Format_(Length,_pages).tsv
(108 total, here are 4)
21198/zz0008z7br
21198/zz0008zj9k
21198/zz0008zjd4
21198/zz0008zjfn

Title.tsv
21198/zz002k6twv
21198/zz002kdmk0
```

Thanks --Andy

export_scripts/export_metadata.py
```diff
@@ -0,0 +1,50 @@
+import oracledb
+import pandas as pd
+import api_keys
+
+query = """
+select
```
akohler, line 6: @kjallen I checked for duplicate ARKs in the Chron file, since we moved that to `ProjectItem.coverage` as a non-repeatable field:
```
$ tail -n +2 Chronological_period_covered.tsv | cut -f 1 | sort | uniq -d
21198/zz0009frh8
$ grep "21198/zz0009frh8" Chronological_period_covered.tsv
21198/zz0009frh8                1950-2000
21198/zz0009frh8                1950-2000
```
Just one repeated ARK, and the value is identical so no loss of data.  But it should be easy to filter out duplicate rows of all types via a `SELECT DISTINCT`, which I think would apply to all of the types, repeatable or not - there can be dup ARKs, but whole rows should all be unique.

akohler:
@kjallen Last week in Slack, talking about the models, and `Relation`:
```
Andy Kohler
  [5:13 PM](https://uclalibrary.slack.com/archives/D0DSG4MJA/p1678414407561789)
There seems to be only one RelationType?  If so, do we need that model at all?

Kristian Allen - UCLA
  [5:14 PM](https://uclalibrary.slack.com/archives/D0DSG4MJA/p1678414475484129)
Good point, if they haven't added an additional type by now, I don't think that would change

Andy Kohler
  [5:14 PM](https://uclalibrary.slack.com/archives/D0DSG4MJA/p1678414495492719)
relatedTo seems implicit in a Relation concept :slightly_smiling_face: - but if we want to be explicit about parent/child/sibling, then having a type seems relevant.

Kristian Allen - UCLA
  [5:20 PM](https://uclalibrary.slack.com/archives/D0DSG4MJA/p1678414803455359)
yeah, this was a hold over for modeling MODS and I don't think was fulled used, other projects use isPartOf etc...
I just checked and there is no data for Relation (for OH) in DLCS, so now i'm leaning towards just removing altogether (edited)
```

You exported 26 rows of data in `Relation.tsv`, all `relatedTo`.  Do we need this data?  We don't have model(s) for it.  We can add, but need to decide
* Repeatable?  No repeated ARKs in that file.
* Is a `RelationType` needed?  Only `relatedTo` values in exported data.

Thanks --Andy

export_scripts/export_metadata.py
kjallen: Confirming modification of production data:

Removed `NULL` entries for the following terms:

- Coverage (7)
- Title (2)
- Format (108)

Confirmed and removed duplicates

- Coverage (1)

kjallen:
I spoke with Lisa about this and she would like to keep this field. For now, I will keep this in the export and have additional ticket for Relation structure.

She will get back to us on the above questions:
- Repeatable? No repeated ARKs in that file.
- Is a `RelationType` needed? Only `relatedTo` values in exported data.

akohler (approved):

akohler:
@kjallen Thanks for making the changes.  I'm approving and merging this, though there's still the open question of `Relation`.  I also found more empty rows in other data files, but since that's likely another data cleanup issue, I'll open a new ticket for that.  For now it's not a blocker.

PR #22, opened 2023-03-17
SYS-1189: Add coverage, status and parent info to item editing screen

akohler:
Implements [SYS-1189](https://jira.library.ucla.edu/browse/SYS-1189).

This PR adds 3 fields to the `edit_item` template:
* coverage (free text, editable)
* status (controlled list, editable)
* parent (link, if parent exists, not editable)

A `get_default_status()` method was added to `models.py` to support the need for "In progress" to be the default status.  I also fixed the existing default type, so it now uses the `get_default_type()` method.

### Testing
* Load full set of project item data, if not already done:
`$ docker-compose exec django python manage.py import_projectitems oh-projectitems-export-3.csv`
* Do a search, like http://127.0.0.1:8000/search_results/title/allen
* View a few records. Interviews and legal agreements generally are children, so there should be a parent link.  Series are top-level and have no parent, so `None` should display.
* Edit coverage / status as desired  and confirm the changes persist when saved.

### DIFF
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -1,10 +1,13 @@
 from django import forms
 from oh_staff_ui.models import (
+    ItemStatus,
     ItemType,
     Name,
     NameType,
     Subject,
     SubjectType,
+    get_default_status,
+    get_default_type,
 )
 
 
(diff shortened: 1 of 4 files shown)


ztucker4 (approved):
Hi @akohler, this looks good. I can view and edit Coverage and Status on items, and changes are correctly saved. Navigation from children to parents also works, and top-level items display Parent as None. Audio makes sense as a default ItemType since it applies to more than half of the current Items.
Merging now, thanks!

PR #23, opened 2023-03-21
SYS-1191: Add remaining controlled item fields to edit form

akohler:
Implements [SYS-1191](https://jira.library.ucla.edu/browse/SYS-1191).

This PR adds the ability to edit the remaining optional metadata fields which are managed via a controlled values list:
* Publisher
* Resource
* Copyright
* Language

`Resource` (type of resource / collection) did not exist, but is present on the legacy DLCS editing form and in the data exported and made available via #21.  I added models for `Resource` (the values), `ResourceType` (the qualifier types), and `ItemResourceUsage` the many-to-many association between items and resources), along with `admin.py` displays for the first two.  I also added fixtures for the first two, automatically loaded with our other fixtures via `entrypoint.sh`.

The code is all similar to previous code for editing `Name` and `Subjects` data.  The 4 new fields follow those 2.  A form with no data looks like this:
sample_1975

### Testing
If running, restart to apply new migration and load new fixture data:
```
$ docker-compose down
$ docker-compose up -d
```
Retrieve an item via search, or create a new one, and try adding / changing / deleting the various fields.


ztucker4 (approved):
Hi @akohler, this looks good. I can edit and save one or more Publishers, Resources, Copyrights, and Languages on items. Adding/editing Resources and ResourceTypes in the admin interface also works.
I find the term "Resource" confusing here, but it does align with the current interface so that seems fine for now. I'm curious if there are any instances of the "genre" resource type in the current data - if not, maybe we can change this to an unqualified field like Language.
Merging now, thanks!

akohler:
@ztucker4 , I agree "Resource" is a little confusing.  In my next PR, for the remaining elements on this form, I'll change the labeling to "Resource Type" to better match the legacy application.

There are instances of "genre" in the exported data, though only 5 of those compared with 2517 "typeOfResource".

PR #24, opened 2023-03-23
SYS-1180: Display hierarchy on item edit pages

ztucker4:
Implements [SYS-1180](https://jira.library.ucla.edu/browse/SYS-1180)

This PR adds a "local hierarchy" display to item edit pages. This hierarchy is presented in a collapsible tree up to the Series level. Users can expand and collapse the tree and use links to navigate to related items.

I've set this as a draft PR for now since I have a few concerns about my solution:
* The recursive item_tree template may be overcomplicated and difficult to maintain.
* We could probably use more styling to highlight what page we're currently on in the hierarchy.
* I'm not sure how far the tree should be "open" by default. Currently, the Series is always expanded, as is the immediate parent of the current item. "Interview" level items are not expanded by default, but displayed alongside siblings.
    * This implementation is also not designed for more than 3 hierarchy levels (series, interview, file) - not sure how flexible we need to be in general.

Screenshots:
Series level
image
Interview level
image
File level
image

### DIFF
--- a/oh_staff_ui/templates/oh_staff_ui/item_tree.html
+++ b/oh_staff_ui/templates/oh_staff_ui/item_tree.html
@@ -1,0 +1,50 @@
+{% for parent, children in relatives.items %}
+    {# if this is topmost parent or immediate parent of page we're on, open the details #}
+    {% if item in children or parent == top_parent %}
+        {% if children %}
+            <li>
+                <details open> 
+                    {# if this is the page we're on, don't link #}
+                    {% if parent == item %}
+                    <summary>{{parent}}</summary>
+                    {% else %}
+                    <summary><a href="{% url 'edit_item' parent.id %}">{{parent}} </a></summary>
+                    {% endif %}
+                    <ul>
+                        {% include "oh_staff_ui/item_tree.html" with relatives=children %}
+                    </ul>
+                </details>
+            </li>
+        {% else %}
+            {# if this is the page we're on, don't link #}
+            {% if parent == item %}
+                <li>{{parent}}</li>
+            {% else %}
+                <li><a href="{% url 'edit_item' parent.id %}">{{parent}}</a></li>
+            {% endif %}
+        {% endif %}
+    {% else %}
+        {% if children %}
+            <li>
+                <details> 
+                    {# if this is the page we're on, don't link #}
+                    {% if parent == item %}
+                    <summary>{{parent}}</summary>
+                    {% else %}
+                    <summary><a href="{% url 'edit_item' parent.id %}">{{parent}} </a></summary>
+                    {% endif %}
+                    <ul>
+                        {% include "oh_staff_ui/item_tree.html" with relatives=children %}
+                    </ul>
+                </details>
+            </li>
+        {% else %}
+            {# if this is the page we're on, don't link #}
+            {% if parent == item %}
+                <li>{{parent}}</li>
+            {% else %}
+                <li><a href="{% url 'edit_item' parent.id %}">{{parent}}</a></li>
+            {% endif %}
+        {% endif %}
+    {% endif %}
+{% endfor %}
--- a/oh_staff_ui/templates/oh_staff_ui/item_tree.html
+++ b/oh_staff_ui/templates/oh_staff_ui/item_tree.html
@@ -1,50 +1,25 @@
 {% for parent, children in relatives.items %}
     {# if this is topmost parent or immediate parent of page we're on, open the details #}
-    {% if item in children or parent == top_parent %}
-        {% if children %}
-            <li>
-                <details open> 
-                    {# if this is the page we're on, don't link #}
-                    {% if parent == item %}
-                    <summary>{{parent}}</summary>
-                    {% else %}
-                    <summary><a href="{% url 'edit_item' parent.id %}">{{parent}} </a></summary>
-                    {% endif %}
-                    <ul>
-                        {% include "oh_staff_ui/item_tree.html" with relatives=children %}
-                    </ul>
-                </details>
-            </li>
-        {% else %}
-            {# if this is the page we're on, don't link #}
-            {% if parent == item %}
-                <li>{{parent}}</li>
-            {% else %}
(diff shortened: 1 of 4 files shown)


oh_staff_ui/views_utils.py
```diff
@@ -93,6 +96,8 @@ def get_edit_item_context(item_id: int) -> dict:
         "publisher_formset": publisher_formset,
         "resource_formset": resource_formset,
         "subject_formset": subject_formset,
+        "relatives": relatives,
+        "top_parent": top_parent,
```
akohler, line 100: Since `relatives` contains `top_parent`, it seems redundant to include `top_parent` explicitly in the context.

oh_staff_ui/views_utils.py
```diff
@@ -84,6 +84,9 @@ def get_edit_item_context(item_id: int) -> dict:
     copyright_formset = get_copyright_formset(item_id)
     language_formset = get_language_formset(item_id)
     resource_formset = get_resource_formset(item_id)
+    top_parent = get_top_parent(item)
```
akohler, line 87: `get_edit_item_context()` doesn't need to know how to assemble what's needed.  I recommend adding a `get_relatives(item)` method which builds `relatives`, which can then be included in the context.  See my other comment about `top_parent` seeming redundant in the context.

oh_staff_ui/templates/oh_staff_ui/item_tree.html
```diff
@@ -0,0 +1,50 @@
...
+        {% else %}
+            {# if this is the page we're on, don't link #}
+            {% if parent == item %}
+                <li>{{parent}}</li>
+            {% else %}
+                <li><a href="{% url 'edit_item' parent.id %}">{{parent}}</a></li>
+            {% endif %}
+        {% endif %}
+    {% else %}
+        {% if children %}
```
akohler, line 27: Lines 27-48 are exactly the same as lines 4-25, except the ` open` on line 6 vs line 29.  I think this could just be one block of code, with the test on line 3 moved to what's currently line 6, something like
```
<details {% if item in children or parent == top_parent %} open {% endif %}>
```

akohler:
@ztucker4 I think this is an excellent approach.  I've added some code-specific comments which should be factored in.

More generally, I'm not sure the template needs to be recursive.  Since the context passed to the `edit_item` template includes (or should include) the full set of data needed for the `<details>` HTML, it seems like we could just build the complete HTML, then `open` it at the current item.  But I have not experimented with this, so I may well be wrong - I'm just referring to https://iamkate.com/code/tree-views/ which says
```
... for each list item that contains a nested list, we put the contents of the list item inside <details> and <summary> 
elements, using the open attribute to control which nested lists are initially expanded
```

Finally, I wrote some code to explore the current data, using your functions.  I confirmed that 2 levels of descendants is all we have (and @kjallen agreed via Slack conversation), and I found that the largest single-series hierarchy we currently have is the massive `Z: Orphan Interviews pre 1999`, with 2386 total descendants... nearly a quarter of all items!

But even that displays in less than a second on my local system:
http://127.0.0.1:8000/item/16

And the largest "real" series has 311 items, and displays nearly instantly: http://127.0.0.1:8000/item/14

So I think the recursive approach is not a performance problem... I just wonder if it can be eliminated, and think that's worth exploring more - and documenting in code, if it turns out to be the best approach.

ztucker4:
Hi @akohler, this is ready for a second look. I've implemented the suggested tweaks to `views_utils` and rewritten `item_tree` to loop over the three hierarchy levels rather than use recursion. 

To help prevent large trees (like the Orphan Interviews ones) taking up the whole screen above the item details form, I've put everything under a top-level "Hierarchy Context" element which is collapsed by default. Since most trees are smaller this might not be needed, but let me know what you think. There's also a tiny bit of new CSS to highlight the current item.

oh_staff_ui/views_utils.py
```diff
@@ -410,3 +412,28 @@ def save_item_subjects(item: ProjectItem, subject_formset_data: list) -> None:
...
+def get_top_parent(item: ProjectItem) -> ProjectItem:
+    current_parent = item
+    while current_parent.parent:
+        current_parent = current_parent.parent
+    return current_parent
+
+
+def get_descendants(item: ProjectItem) -> dict:
+    descendants = {}
+    children = ProjectItem.objects.filter(parent=item).order_by("sequence", "title")
```
akohler, line 426: Good catch! I had not noticed the unsortedness.

akohler:
@ztucker4 Here's what I (thought I) commented about last night - two UX-related issues I think are not blockers, but should be improved at this point if possible / practical.  We'll also collect feedback from OH staff soon, so we could leave these as-is for now, but I suspect they would note these issues right away.

1) The top-level "Hierarchy Context".  I suggest replacing that with the Series title.  However, would that cause challenges with the open/closed logic?  I like having this closed by default, especially in its current position.  Moving it into a left panel / frame could allow it to be open by default, but would introduce a different set of UX issues.  If a top-level "container" like this is the best solution for now, that's OK - maybe change the wording to "View series context" or "View children" or something less jargony?

2) Interviews with no children.  I had an example but don't want to revert my WIP database to be able to run this branch again - I think it was http://127.0.0.1:8000/item/325 , looks like that interview has no children.  The "disclosure arrow" appears, but nothing appears when clicked, making it look like something's not working.

Two options might be:
1.  Change the styling to hide the arrow when there are no children.
2.  Display some text when clicked, maybe "No files for this interview" or similar.

If you have other thoughts, feel free - just something to help users understand that all is OK, there's really no data to look at.  This would apply to Series with no Interviews as well, which may just be when setting up a new Series.

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for (hopefully) final review. The latest commit addresses the two UX issues you identified. The Series is now used as the top-level hierarchy container, and it's collapsed by default. Interviews and Series without any children now correctly display as non-expandable list items.

akohler (approved):
@ztucker4 This looks great.  The conditional `files` display works as needed, and I like the removal of the generic hierarchy container.  This is ready to move forward - merging!

PR #27, opened 2023-03-27
SYS-1198: Enable editing of remaining uncontrolled metadata fields

akohler:
Implements [SYS-1198](https://jira.library.ucla.edu/browse/SYS-1198).

TL;DR: This PR allows the editing of "uncontrolled" optional metadata fields - those which are not backed by a list of controlled values.  Staff can create/change/delete these directly via the item editing form.  Most of these are repeatable, so like the "controlled" fields (name, subject etc.), these are backed by Django formsets, allowing users to add as many of these as needed:
1. `Alt ID`
2. `Alt Title`
3. `Description`
4. `Date`
5. `Language`
6. `Format`

1-4 have lists of qualifier types; 5-6 do not.

Details, and other changes:
* There was a lot of repetitive code for these and the "controlled" metadata fields, to create & populate the formsets being passed to the template and to save the form data returned from the template.  This PR adds two methods which now handle that data in generic ways:
  * `get_metadata_formset()`
  * `save_item_formset_data()`
* To support this generic approach, the relevant models now consistently have a generic `value` field (e.g., `Language.value`, `Description.value`, `ItemNameUsage.value`, etc.).  Given the ways these models are used, I feel the loss of specific attribute names is more than offset by the ability to write more generic code against them.
* `Relation` was added as an optional text field to the main `ProjectItem` model & form.
* Updating models requires migrations.  Most of these models have been updated multiple times, leading to a cumulative series of migrations which at times did not play nicely together, so I squashed all migrations into one clean one, `0001_0001_super_squash_everything`, which now has only `CreateModel` operations.  Git(Hub) does a poor job of showing these changes.
* Old migration files have been deleted.
* Some minor changes were needed to `admin.py` and fixtures for `Language` cleanup, and to the `import_names` management command, due to model updates.
* Field display in the `edit_item` template tries to group similar UI elements together.  I expect this will change with OH staff input.
* Minor CSS tweak to align table cells.
* FYI, there were multiple commits as I built & rebuilt this; I squashed them before pushing and creating this PR.

### Testing
As noted above, there are significant model changes and some minor fixture / data loading changes.  Best to take a "clean database" approach:
* Check out this branch: `git checkout SYS-1198/edit_item_uncontrolled_metadata`
* Clean and reload: `docker_scripts/nuke_dev_db.sh`
* Test as needed
* When done testing, check out your current branch and run the `nuke` script again.

For the testing, create a new item or pick any existing one - I've been using http://127.0.0.1:8000/item/331 .

Experiment with adding, changing, and deleting fields of all types.  Combine changes as much as desired.  We can't yet add multiple fields of the same type at once (e.g., add two new names at once), but we can add different fields at once (e.g., add a name, subject, description, format, and date, then click `Save`).


ztucker4 (approved):
Hi @akohler, this looks great. I can add, view, edit, and delete metadata in the new fields (and existing ones) in various combinations. The generic formset-related methods are much nicer to deal with, and the model changes supporting them make sense. Merging now, thanks!

PR #28, opened 2023-03-30
SYS-1200: Initial version of Helm chart

akohler:
Implements [SYS-1200](https://jira.library.ucla.edu/browse/SYS-1200).

This PR adds an initial version of the Helm chart for the Oral History Staff UI application, `oh-staff`.  It's intended for deployment in our production Rancher Kubernetes environment.

There are two external secrets defined.  Both have been created in our AWS parameter store.

This application will need access to the same persistent volumes as the existing [`dlcs-staff-ui`](https://github.com/UCLALibrary/dlcs-staff-ui/tree/main/charts) application, so I copied those templates as-is, changing just the application name/variables.  This application will also need similar resources allocated, so I also used that data.

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -1,0 +1,116 @@
+# Values for oh-staff.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: uclalibrary/oral-history-staff-ui
+  tag: v0.0.1
+  pullPolicy: Always
+
+nameOverride: ""
+
+fullnameOverride: ""
+
+service:
+  type: NodePort
+  port: 80
+
+ingress:
+  enabled: "true"
+  annotations:
+    kubernetes.io/ingress.class: "nginx"
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "600"
+    cert-manager.io/cluster-issuer: sectigo-acme-clusterissuer
+    kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: 'oh-staff.library.ucla.edu'
+      paths:
+        - "/"
+  tls:
+  - secretName: oh-staff-tls
+    hosts:
+      - oh-staff.library.ucla.edu
+
+django:
+  env:
+    run_env: "prod"
+    debug: "false"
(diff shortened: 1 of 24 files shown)


charts/prod-ohstaff-values.yaml
```diff
@@ -0,0 +1,116 @@
...
+image:
+  repository: uclalibrary/oral-history-staff-ui
+  tag: v0.0.1
+  pullPolicy: Always
+
+nameOverride: ""
+
+fullnameOverride: ""
+
+service:
```
sgurnick, line 16: Setting the Service type to `NodePort` is something I had been doing for a while. Recently I realized this is unnecessary for how we are funneling traffic into the cluster. I'll explain this more next time we meet. 

Let's use the `ClusterIP` Service type for this app (and most likely all others moving forward). You would just change this stanza to the following:
```
service:
  type: ClusterIP
```

sgurnick: Yes - it should be changed in the default `values.yaml` file as well.

sgurnick (commented):
@akohler Other than my comment about the Service type, I don't immediately see anything else that needs adjusting.

I do have questions about how this app will co-exist with the existing dlcs-staff-ui app that we have deployed. Since this new oh-staff-ui app is mounting the same NFS shares as persistent volumes, we want to be careful with any conflicts. This may warrant a further discussion, but I wanted to mention that as something to be aware of.

charts/prod-ohstaff-values.yaml
akohler: Changed  both files.

akohler:
> I do have questions about how this app will co-exist with the existing dlcs-staff-ui app that we have deployed. Since this new oh-staff-ui app is mounting the same NFS shares as persistent volumes, we want to be careful with any conflicts. This may warrant a further discussion, but I wanted to mention that as something to be aware of.

This will eventually replace the `dlcs-staff-ui` app - but they do need to co-exist for a while.  If we have "test" (or similar) versions of those NFS shares, it would be better to mount those into this application for now... but I suspect we do not.  We could set up testing directories within those shares/volumes.  Or we (someone with access, at least) could create temporary fake shares for this application to avoid conflicts until we're closer to cutover.

akohler:
@sgurnick I've updated the volume paths as we discussed today - please see 541d12f .  If all looks OK now, please approve and try a deployment.

sgurnick (approved):

PR #31, opened 2023-04-06
SYS-1196: Remaning metadata loading commands

ztucker4:
Implements [SYS-1196](https://jira.library.ucla.edu/browse/SYS-1196)

This PR adds new Django management commands and updates the existing import_names to work with the updated TSV files. Each command takes a single argument: the path to the desired import file, relative to the top directory. For example, `docker-compose exec django python manage.py import_descriptions export_scripts/Description.tsv`

To support longer Descriptions, this PR also includes a change to the Descriptions model and an associated migration, which needs to be run before any testing.

Full list of import_* commands:

* import_altids
* import_alttitles
* import_copyrights
* import_coverages
* import_dates
* import_descriptions
* import_formats
* import_languages
* import_names
* import_publishers
* import_relations
* import_resources
* import_subjects

Data rows will be rejected for import if the Value column is blank, or if the associated Type (e.g. NameType) or AuthoritySource does not exist. Rejected rows are printed to the console and a total count of rejected rows is printed at the end.

Testing notes: there are some CopyrightTypes ("permission" and "rightsHolderNote") in the provided Copyright data that should possibly be added to the fixtures. There is also one unusual SubjectType in the Subject data that might indicate a problem with that export ("Jones, Marcia : local").


oh_staff_ui/management/commands/_import_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def get_dicts_from_tsv(filepath=str) -> list:
+    with open(filepath, mode="r") as f:
+        dict_reader = DictReader(f, delimiter="\t")
+        full_dicts = list(dict_reader)
+    return full_dicts
+
+
+def get_or_create_name(source=AuthoritySource, value=str) -> Name:
```
akohler, line 20: Since these routines all do the same thing, they could be combined into one generic one, similar to [`save_item_formset_data`](https://github.com/UCLALibrary/oral-history-staff-ui/blob/main/oh_staff_ui/views_utils.py#L257).  However, given the limited usage of these, we can keep these as-is.

akohler (approved):
@ztucker4 These are all good: i ran all, after applying the migration, and got no errors.  I confirm the various data problems you noted and will add those to [SYS-1195](https://jira.library.ucla.edu/browse/SYS-1195) for @kjallen to consider.

I did push a commit with some minor tweaks to other files:
* modified `import_projectitems.py` to remove the hard-coded `fixtures` requirement
* updated `nuke_dev_db.sh` so the optional data reloading includes all of these, for convenience

Merging now, thanks --Andy

PR #33, opened 2023-04-06
SYS-1186: Fix error when searching for full ARK

ztucker4:
Implements [SYS-1186](https://jira.library.ucla.edu/browse/SYS-1186)

Changing the URL converter from `str` to `path` now allows search terms including forward slashes. Useful for full ARK search and title searches including dates (e.g. 11/18/2020).


akohler (approved):
@ztucker4 Excellent, elegant fix!  Tested with full and partial ARKs, and with title searches including full & partial dates, all works perfectly.  Merging --Andy

PR #36, opened 2023-04-07
SYS-1212: Allow parent to be specified when adding new records

ztucker4:
Implements [SYS-1212](https://jira.library.ucla.edu/browse/SYS-1212)

These changes allow any ProjectItem with Type "Series" or "Interview" to be selected as the parent when adding a new item. This is done with a dropdown field at the top of the add_item form. 
Additionally, a new link called "Add a child item" has been added to the top of the item display page for Series and Interviews. Clicking this link should take the user to the add_item form with the Parent dropdown already set. 

This does not allow moving children to new parents at this point.

### DIFF
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -11,6 +11,7 @@ from oh_staff_ui.models import (
     Language,
     Name,
     NameType,
+    ProjectItem,
     Publisher,
     PublisherType,
     Resource,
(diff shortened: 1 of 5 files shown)


akohler (changes requested):
@ztucker4 Overall this is excellent!  As implemented, I have just one technical change requested, to simplify the filter on `ProjectItemForm.parent`.

I'm uncertain about allowing editors to move items to a new parent via the `parent` drop-down currently showing on the edit page.  Part of me feels like this could be dangerous... but the rest of me thinks this would be the simplest solution, fitting neatly into the rest of the design.  

So for now, please enable this on the editing form, setting `parent`.  If a user selects a new parent for the item being edited, save their choice by enabling it in `save_all_item_data()`.

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for re-review. The Parent dropdown in the edit_item context can now be edited and saved (along with all other basic item metadata). Thanks for the tip on `__in` for filters too!

akohler (approved):

akohler:
@ztucker4 - Confirmed that I can now update parents.  There currently are no restrictions: I can move whole series under new parents, or move interviews or file-level items to have no parent.  We may want to add some logic later to prevent these, but it's so easy to move things around that fixing incorrect moves is just as easy for editors to do... so this is fine with me for now.

As expected, moving an item with children retains those connections, just changing the parent pointer of the item in question.  The hierarchy display updates as it should.

Excellent work! - merging.

PR #39, opened 2023-04-12
SYS-1204: Enable dynamic repeatable fields

akohler:
Implements [SYS-1204](https://jira.library.ucla.edu/browse/SYS-1204).

This PR adds dynamic repeatable fields to the item editing form.  It allows the editor to create any number of names, subjects, descriptions etc., without having to save and submit the form after each one.  It does this by using javascript to clone a hidden Django `formset.empty_form`.  The javascript updates the DOM and the hidden `TOTAL_FORMS` value which Django uses to confirm the correct number of sets of form data was received.

The `edit_item` template now has an extra row for each metadata type, like `<tr id="name_empty_form" class="empty_form">`.  This is largely duplicate code, but I don't see a way to output both the "real" form data for display and the hidden empty form, using Django's template language, without duplicate code.  `formset.empty_form` is not a form itself, so can't be output when iterating over `for form in formset`.

Bumps version to `v0.3.0`.

Testing:
* No migrations required.
* Create a new item, or retrieve any existing one.
* Use the `+` buttons next to each metadata type to add any number of new form elements.
* Fill out some / all new forms.
* Combine with editing existing form data, deleting some as well.  
* Try as many different combinations as possible.
* After clicking `Save`, confirm data was added / updated / deleted as requested.

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v0.2.0
+  tag: v0.3.0
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/templates/oh_staff_ui/base.html
+++ b/oh_staff_ui/templates/oh_staff_ui/base.html
@@ -3,7 +3,10 @@
 <html>
 
 <head>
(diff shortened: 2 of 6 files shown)


ztucker4 (approved):
This all looks good to me. Testing (adding/updating/deleting data in various combinations) works correctly. I also don't see any immediate way to improve the duplicate template code since the empty_forms are not in the formsets being looped over.
Merging, thanks!

PR #42, opened 2023-04-14
SYS-1227: Add initial support for files

akohler:
Implements [SYS-1227](https://jira.library.ucla.edu/browse/SYS-1227).

This PR adds 2 models which support the association of media files with items:
* `MediaFileType`: A lookup model, equivalent to the legacy `file_groups` table.
  * This model is visible in the Admin interface.
  * `oh_staff_ui/fixtures/media-file-type-data.json` provides the 15 values for this model, copied from the legacy database.
* `MediaFile`: A data model, corresponding to the legacy `content_files` table.
  * This supports 1-to-many relation with `ProjectItem` (an item can have any number of files).
  * This model is self-referential, supporting parent / children.  For example, a "master" wav audio file can have "submaster" mp3 file(s), or a "master" tiff image file can have various jpg derivative files.
  * This should make it easier to handle sets of files consistently.  For example, deleting a "master" audio file could also delete its submasters, while deleting a submaster would not require deletion of the master.
  * This model only has metadata which can't be derived easily from other data.  It omits legacy data like
    * mime type / content type / file use: get these from the associated `MediaFileType`
    * size: available from the underlying Django `File` object
    * location / url: Django provides both path on disk and addressable URL

Django `FileField`s require an `upload_to` parameter which specifies the target directory.  This can be a constant value, or a callable which sets the value for each instance.  Since files need to be saved into various directories based on types / uses, I've set this to a callable, `get_file_directory()`.  For now this just uses the `samples/managed/` directory; more complex logic will be added later.

Testing:
* Restart local application to apply the migration and to load the new fixture.
* Run tests (yes, actual tests!) to confirm basic functionality
* Review test code in `oh_staff_ui/tests.py` for examples of working with the new models.
```
# If running
$ docker-compose down

$ docker-compose up -d
$ docker-compose exec django python manage.py test

# Expected output
Found 2 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
..
----------------------------------------------------------------------
Ran 2 tests in 0.038s

OK
Destroying test database for alias 'default'...
```


kjallen (approved):
Deleted volumes and containers, rebuilt, and ran tests

PR #43, opened 2023-04-14
SYS-1225: Add uniqueness constraints to metadata models

ztucker4:
Implements [SYS-1225](https://jira.library.ucla.edu/browse/SYS-1225)

This PR adds uniqueness constraints to the "item usage" models with limited values. That is, the combination of Item, Type, and Value must be unique for the following ProjectItem metadata types: Name, Subject, Publisher, Copyright, and Resource. Language usage must be unique by Item and Value.

To test, apply the new migration and try adding duplicate metadata. If you attempt to add a row that duplicates another, you should see an ugly IntegrityError page on form submission and your changes will not be saved (including any potentially valid updates you attempted at the same time). 

Uniqueness constraints could also be added for fields with free text entry (e.g. Description), but accidental duplication is less probable for these. We may also want to enforce uniqueness for the metadata objects themselves (e.g. disallow creation of two Statuses with the same name or description) - let me know if these should be added to this PR.


oh_staff_ui/models.py
```diff
@@ -116,6 +116,17 @@ class ItemLanguageUsage(models.Model):
         Language, on_delete=models.PROTECT, blank=False, null=False
     )
 
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=[
+                    "item",
+                    "value",
```
akohler, line 124: Minor nitpick: remove the trailing comma after "value", and on line 126 after the constraint name, to be consistent with the other new constraints (e.g., see "name_value_type_unique" on line 170).  Or, change all the other new constraints to match this one, but for these short lists I prefer the compact format.

akohler (changes requested):
@ztucker4 Other than one extremely minor formatting issue, the current code is fine.  However, I'd like you to add two enhancements to this implementation.

1) Instead of showing an uncaught exception (which won't show the same way in production, with debug turned off), let's be friendlier.  I think these potential duplicates can be caught via javascript: iterate over the relevant types, and if a pair of inputs has the same values, show a message with the type & value.  This probably should be a second function; move the code from `validateEditItemForm()` into a separate function - `checkMissingValues()` or similar - and then call both functions from `validateEditItemForm()`.

The alternative is catching the Python exception in `save_all_item_data()`, which would require passing messages back to the form for display; extending the javascript approach seems much cleaner.

2) Please add some test code, in `tests.py`.  Testing these constraints isn't mandatory... but it's useful to have tests for them which don't require the UI.  And I want our team to incorporate more test code into projects, so this is a good case to practice with.

This can be one test case - `MetadataUniquenessTestCase` or similar - with multiple tests, one for each relevant metadata type.  I've just written one of these as practice for myself; please let me know if you'd like to see that as an example, or if you have any questions.

As for your question: I agree, no need to enforce uniqueness constraints on free text fields.  And lookup models like `ItemStatus` and `MediaFileType` are short, simple, and managed by editors via the Admin interface, so I think we don't need to add these constraints to those.

Thanks --Andy

ztucker4:
Hi @akohler , this is ready for re-review. I've added tests for each of the newly constrained models (thank you for the example!), added JS validation to check for duplicate type/value pairs, and made the requested formatting fix.

akohler (approved):

akohler:
@ztucker4 This looks great!  Confirmed all tests pass (and they look correct :) ), and the javascript does what it should.  Tested manually via editing form and always got the results I expected.

We might consider a future enhancement of checking whether the "Delete" box is checked.  For example, I tried to create duplicate subject; got alerted; decided I didn't want to add another subject, so checked the Delete box without editing the values; still got alerted.  But for now this is fine, and I'm merging.  Thanks --Andy

PR #45, opened 2023-04-18
SYS-1228: Implement basic file saving

akohler:
Implements [SYS-1228](https://jira.library.ucla.edu/browse/SYS-1228); bumps version to `v0.5.0`.

This PR adds the ability for editors to select and "upload" files, associating them with items.  Files are not really uploaded (transferred from local machine via browser & network to server); instead, the user selects a file from a hard-coded directory on the server itself.  This reuses functionality from the phase 1 project.

For now, files are selected via form at `/upload_file/<item_id>`.  Users can switch between file and metadata forms by clicking `View metadata` or `View files`.

This does not implement file processing (converting audio wav files to mp3, or image tiffs to jpeg, etc.).  For now, the selected file is simply copied to the `samples/managed/` subdirectory, renamed, and a new `MediaFile` for that new file is associated with the current `ProjectItem`.  Proper file handling will be implemented in later tickets.

Known problems:
* New file names currently are based solely on item ARK and the selected file name.  If Django sees that a file with the same new name already exists in `samples/managed/`, it will make another copy with a new, Django-modified, name.  For example, if `samples/managed/21198-zz000906d5_fake_file_01.wav` already exists, Django will create a new file like `samples/managed/21198-zz000906d5_fake_file_01_XT5EfmP.wav` (or some other random characters appended).  This will be addressed later, once we talk about how to handle this...

Testing:
* Select any item
* Click `View files` to see the file upload form.
* Use the form to add files, one at a time.
  * As files are added, they display in a table above the form.
* If desired, review the data

```
# Added 3 files via UI form
$ ls -al samples/managed
-rw-r--r-- 1 akohler akohler    13 Apr 18 13:50 21198-zz000906d5_fake_file_01.wav
-rw-r--r-- 1 akohler akohler  2154 Apr 18 13:51 21198-zz000906d5_sample.pdf
-rw-r--r-- 1 akohler akohler 44144 Apr 18 13:51 21198-zz000906d5_sample.wav
-rw-r--r-- 1 akohler akohler    71 Apr 18 13:27 .gitignore

$ docker-compose exec django python manage.py shell
>>> from oh_staff_ui.models import MediaFile
>>> MediaFile.objects.all()
<QuerySet [<MediaFile: MediaFile object (11)>, <MediaFile: MediaFile object (12)>, <MediaFile: MediaFile object (13)>]>

>>> for f in MediaFile.objects.all():
...   # Delete file from disk
...   f.file.delete()
...   # Delete MediaFile itself
...   f.delete()
...
(1, {'oh_staff_ui.MediaFile': 1})
(1, {'oh_staff_ui.MediaFile': 1})
(1, {'oh_staff_ui.MediaFile': 1})

>>> MediaFile.objects.all()
<QuerySet []>

>>> exit()

# Files have been deleted
$ ls -al samples/managed
-rw-r--r-- 1 akohler akohler    71 Apr 18 13:27 .gitignore
```

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v0.4.5
+  tag: v0.5.0
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -1,4 +1,6 @@
+import os
 from django import forms
+from django.core.cache import cache
 from oh_staff_ui.models import (
     AltIdType,
     AltTitleType,
@@ -9,6 +11,7 @@ from oh_staff_ui.models import (
     ItemStatus,
     ItemType,
     Language,
+    MediaFileType,
     Name,
     NameType,
     ProjectItem,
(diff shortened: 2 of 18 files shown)


kjallen:
This is way edge scenario, but just noting I do get error in following (unlikely) scenario

- Load file upload page
- Open shell, rename file you will attempt to upload, or change permissions to unreadable
- Select and attempt to upload the original file name

akohler:
> This is way edge scenario, but just noting I do get error in following (unlikely) scenario
>     * Load file upload page
>     * Open shell, rename file you will attempt to upload, or change permissions to unreadable
>     * Select and attempt to upload the original file name

@kjallen I think this is pretty unlikely.  Files can be uploaded only from the given directory (and its sub-directories), which in production is a Windows share mounted on the server.  The mount defines permissions so that the server should always be able to read what's in it.  The Helm chart defines this at https://github.com/UCLALibrary/oral-history-staff-ui/blob/main/charts/prod-ohstaff-values.yaml#L72-L76 - it's read-only from the server's POV, though OH staff can make changes from the Windows side - but whatever they do from that end, should always be readable from the server's end.

The Django form caches the list of files for 5 seconds (interval definable by us).  I think that list would only get updated if the HTML page is refreshed, so it is possible that a file could be renamed or deleted, and the list would be stale when the form is submitted.  Given the size of the OH staff and the nature of their work, I think it's unlikely that would happen.  

In general though, your point is valid: we should add code which catches and handles file errors.  I'll keep that in mind as I work on the file processing tickets.

kjallen (approved):
Reviewed files, confirmed wav, pdf, txt uploads get properly moved and renamed as per description

PR #46, opened 2023-04-21
Sys 1230/project item export update

kjallen:
Some changes to import process for constancy:

- Script added for `ProjectItem` export source data in export directory
- `coverage` and `relation` added to export source data
- `coverage` and `relation` added to `import_projectitems.py`
- `import_projectitems.py` updated for tabs, sequence casting and new date mask
-  Original csv deleted, new tsv moved to export directory for consistency
- `nuke_dev_db.sh` and `reset_prod_db.sh` updated to reflect new paths
- pk integrity test added for ProjectItems, though ark should be unique as well, we do not enforce in db
- Field type updates for `create_date` and `last_modified_date`

To test:

- Start containers, run `nuke_dev_db.sh` to drop DB
- Run `docker-compose exec django python manage.py import_projectitems export_scripts/project-items-export.tsv`

To verify new date format:
`docker-compose exec django python manage.py shell`
```
(InteractiveConsole)
>>> from oh_staff_ui.models import ProjectItem
>>> p = ProjectItem.objects.get(pk=31)
>>> p.last_modified_date
datetime.datetime(2015, 9, 1, 18, 37, 51, tzinfo=datetime.timezone.utc)
```
* Your pk may be different object depending on load order

Project items should be imported without error. A warning message be presented for ark `21198/zz0008zh0f`, this is the original root/collection object and is skipped when adding items.


oh_staff_ui/management/commands/import_projectitems.py
```diff
@@ -82,12 +84,12 @@ def add_projectitem(self, pi_dict: dict, has_parent=True) -> list:
         p.save()
 
     def format_date(self, date: str) -> str:
-        return datetime.datetime.strptime(date, "%d-%b-%y").strftime("%Y-%m-%d")
+        return datetime.datetime.strptime(date, "%Y-%m-%d %H:%M:%S").strftime("%Y-%m-%d")
```
akohler, line 87: The additional time specification is unnecessary, as the model defines `ProjectItem.create_date` and `ProjectItem.last_modified_date` as `models.DateField` - any time detail is lost:
```
>>> ProjectItem.objects.get(pk=10259).create_date
datetime.date(2022, 9, 6)
```
If this is a concern, we should revise the model to use `DateTimeField` instead.

README.md
```diff
@@ -80,7 +80,7 @@ The container runs via `docker_scripts/entrypoint.sh`, which
    $ docker-compose exec django python manage.py loaddata --app oh_staff_ui seed-data
    #
    # Load full set of item data
-   $ docker-compose exec django python manage.py import_projectitems oh-projectitems-export-3.csv
+   $ docker-compose exec django python manage.py import_projectitems project-items-export.tsv
```
akohler, line 83: This example should have the path in front of the filename: `export_scripts/project-items-export.tsv`

docker_scripts/nuke_dev_db.sh
akohler, line 1: Since `import_projectitems` now handles coverage and relations, the commands to load those should be removed from this script.

docker_scripts/reset_prod_db.sh
akohler, line 1: Since `import_projectitems` now handles coverage and relations, the commands to load those should be removed from this script.

akohler (changes requested):
@kjallen In addition to the minor changes noted in comments, please also delete the `import_coverages.py` and `import_relations.py` management commands, and the corresponding data files in the `export_scripts` directory.  Those are redundant now.  You can also remove / update your export process for these two separate data types.

Once those changes are made, I'll approve & merge this.

I commented in Slack that we should resolve the `null` vs `blank` issue for `coverage` and `relation`.  If you agree with my recommendation - `Change the model to allow only one of blank / null, instead of both as we currently have it.` - that model change, and the migration it requires, should be added to this PR.  If we need to discuss, that work could be done later via a new ticket.

Thanks --Andy

oh_staff_ui/management/commands/import_projectitems.py
kjallen: Updated DB field to `DateTimeField`. Form display remains the same for these fields, this level of precision on the display side is not required.

README.md
kjallen: Updated `README.md` to properly reflect path

akohler (approved):

PR #47, opened 2023-04-21
SYS-1235: Apply default sorts to data in Admin interface

ztucker4:
Implements [SYS-1235](https://jira.library.ucla.edu/browse/SYS-1235)

All lookup tables and controlled value lists are now sorted alphabetically by default within the admin interface. When more than one column is visible, data is sorted in order of the columns left to right.

No migrations or other special testing instructions - just log in to the admin interface and view the data for each available model.


akohler (approved):
@ztucker4 Looks OK to me - merging.

PR #48, opened 2023-04-21
SYS-1236: Create editor user group

ztucker4:
Implements [SYS-1236](https://jira.library.ucla.edu/browse/SYS-1236)

This PR adds a fixture for a new "Oral History Staff" user group. This group has full permissions to view, edit, add, and delete data of all types exposed in the admin interface, but no access to Users, Permissions, or other data. Users in this group are able to use the regular (non-admin) UI to view, add, and edit ProjectItems.


akohler (changes requested):
@ztucker4 This works as expected: Once I loaded the fixture and created a non-admin user for myself (and checked the "Staff" box to get admin access), I was able to log in, interact with the models listed in Admin, edit via the regular UI, confirmed I don't have access to full admin functions like users.

However, there's some inconsistency in what the group includes.  It does include all 19 of the models exposed via `admin.py`, which is good.  But the group includes 4 other models:
```
"delete_altid",
"delete_alttitle",
"delete_date",
"delete_description",
```

If the group only needs to include models exposed in `admin.py`, these 4 should be removed.  If it needs to include all `oh_staff_ui` models, there are many others which are not currently in the group (like `format`, `projectitem`, and all the usage models).

Can you please review and update as needed?  Thanks --Andy

ztucker4:
Hi @akohler, I've updated the fixture and this is ready for re-review. The four extra permission types have been removed, and users can still edit appropriately in the admin and main interfaces. Thanks!

akohler (approved):

akohler:
@ztucker4 Re-tested, all seems fine - merging.

PR #49, opened 2023-04-24
SYS-1239: Prevent duplicate uploads

akohler:
Implements [SYS-1239](https://jira.library.ucla.edu/browse/SYS-1239).

This PR prevents the same file from being associated with a given item multiple times.  If an editor tries, Django does not save the new `MediaFile`, and shows a message like this:
oh_duplicate_file_message

A new test was added which covers this.

Other changes include minor improvements to the file upload form display, and some code formatting changes and other minor cleanup.

Testing:
* Add a file to an item.
* Try to add the same file again to the item.  It should not be added, and a message to that effect should appear.


ztucker4:
Hi @akohler, this all looks good to me. Confirmed that I can't add duplicate files through the upload form, and the new test makes sense and passes. I like the additional labels on the file upload form too. Merging now, thanks!

ztucker4 (approved):

PR #51, opened 2023-04-26
SYS-1240: Improve "edit item" page appearance

ztucker4:
Implements [SYS-1240](https://jira.library.ucla.edu/browse/SYS-1240)

Template, form, and CSS changes intended to give the edit item pages a more cohesive and comprehensible appearance. No functionality changes.

Please test in a couple of different browsers if you can. There are a number of differences between the page's appearance in Firefox and Chrome as it stands - let me know if I should better harmonize these. I've attempted to size things to comfortably fit items with long titles (like 21198/zz002jxz6f).

### DIFF
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -39,7 +39,7 @@ class ProjectItemForm(forms.Form):
         ).order_by("title"),
     )
     title = forms.CharField(
-        required=True, max_length=256, widget=forms.TextInput(attrs={"size": 80})
+        required=True, max_length=256, widget=forms.TextInput(attrs={"size": 140})
     )
     type = forms.ModelChoiceField(
         required=True,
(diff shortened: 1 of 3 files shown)


akohler (approved):
@ztucker4 This looks great! I tested on Windows 10 in Firefox and Edge - I don't have Chrome installed, but current Edge uses Chrome's rendering engine so should be similar enough.

The overall display is much cleaner than before.  Alignment of textareas with textboxes isn't perfect, but it seems to be within a couple of pixels, which is close enough for me.  The grouping of similar inputs looks good; moving Language to the end helps.

I like the Item Information box, especially the "Item Actions" - very helpful.  The new wording "Item Hierarchy Context (use arrow to expand)" is also good.

On my system, this looks better in Firefox, possibly because I'm using a Dark theme in FF.  I only use Edge for comparisons and haven't tried to configure a theme there.  Screenshots below.  I don't know what's involved in making a site respect browser (or OS) theming, but most sites don't, so we're as good as most of the internet now :)

Well done! Merging as is.

In Edge:
oh_edge

In Firefox:
oh_firefox

PR #53, opened 2023-04-27
SYS-1241: Improve links menu and admin navigation

ztucker4:
Implements [SYS-1241](https://jira.library.ucla.edu/browse/SYS-1241)

This PR styles the menu links on each page of the site as a navbar, and adds an additional link from the admin interface to the main site. I'd particularly appreciate feedback about the wording of the links - right now I'm using "Admin (Metadata Configuration)" to link to the admin interface, and "Main Oral History Staff Interface" to link back to the main site. 

Also, there is a preexisting link to the site in the admin interface header ("view site"). I think it's useful to have the link displayed more prominently and with better wording even if it's redundant, but let me know if I should remove the old one. 

Main site header:
image
Admin header:
image

### DIFF
--- a/oh_staff_ui/templates/oh_staff_ui/base.html
+++ b/oh_staff_ui/templates/oh_staff_ui/base.html
@@ -11,7 +11,13 @@
 </head>
 
 <body>
-    <p>Temporary: <a href="/item_search/">Search</a>&nbsp;<a href="/admin/">Admin</a>&nbsp;<a href="/logs/">Logs</a>&nbsp;<a href="/admin/logout/">Logout</a></p>
+    <ul class="navbar">
+        <li><a href="/item_search/">Search (by ARK or Title)</a></li>
+        <li><a href="/admin/">Admin (Metadata Configuration)</a></li>
+        <li><a href="/logs/">Logs</a></li>
+        <li><a href="/admin/logout/">Logout</a></li>
+    </ul>
+    <br>
     {% block content %}
     {% endblock %}
 </body>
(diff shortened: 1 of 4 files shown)


oh_staff_ui/templates/oh_staff_ui/base.html
```diff
@@ -11,7 +11,14 @@
 </head>
 
 <body>
-    <p>Temporary: <a href="/item_search/">Search</a>&nbsp;<a href="/admin/">Admin</a>&nbsp;<a href="/logs/">Logs</a>&nbsp;<a href="/admin/logout/">Logout</a></p>
+    <ul class="navbar">
+        <li><a href="/item_search/">Search (by ARK or Title)</a></li>
+        <li><a href="/add_item/">Add Item</a></li>
```
akohler, line 16: I added a link to `add_item`, which was the default URL route (which I also changed).

oh_staff_ui/urls.py
```diff
@@ -5,7 +5,7 @@
     path("add_item/", views.add_item, name="add_item"),
     path("add_item/<int:parent_id>", views.add_item, name="add_item_from_parent"),
     path("item/<int:item_id>", views.edit_item, name="edit_item"),
-    path("", views.add_item),  # TODO: Change this to something better.....
```
akohler, line 8: I changed the default route from `add_item` to `item_search`, which I'd meant to do for a while.  I expect users normally will want to start with a search when they log in, or switch from Admin to the main application.

akohler (approved):
@ztucker4 This looks great!  I think the wording is fine, and it's OK to leave the default "View site" link in the admin header; the wording you added is clearer and more prominent.

I snuck in 2 minor changes: making the default route (URL) `item_search` instead of `add_item`, and adding an "Add Item" menu link since it was awkward to get to that function otherwise.  Merging, thanks --Andy

PR #54, opened 2023-04-28
SYS-1251: Convert file_utils to OralHistoryFile class

akohler:
Implements [SYS-1251](https://jira.library.ucla.edu/browse/SYS-1251).

This PR moves all functionality from `file_utils.py` (added in #52) into a new class, `oh_staff_ui.classes.OralHistoryFile`.  The methods all were exclusively related to handling uploaded files for this project, and this change will make it easier to handle submasters and thumbnails when that functionality is added.

`OralHistoryFile` generates some data during initialization, instead of everything being done in the `process_media_file()` method.  In particular:
* `content_type`: audio, image, pdf, text
* `target_dir`: the environment-aware directory into which the new file will be written
* `new_file_name`: the new name, based on item ARK, sequence, file use (master, submaster, thumbnail), and file extension

`content_type` will be used to determine how to generate & handle derivatives.  The others are exposed mainly to support testing.

All relevant tests were rewritten to use objects of the new class.  There currently are 37 tests, though 2 of them are being skipped as they require proper submaster handling to be meaningful (the previous versions of those tests were somewhat arbitrary).

The `process_file` management command was updated to use the new class.  Functionality was tested via command-line and via the UI, with everything working as expected.  There are some TODO/WIP comments which will be removed when derivative processing is added.


kjallen (approved):
- Reviewed `OralHistoryFile.py`, `process_file.py` and `tests.py`
- Ran automated tests
- Tested UI, adding/removing files to new and existing items. Confirmed local file locations
- Tested management command, adding files to existing items. Confirmed local file locations

PR #55, opened 2023-05-01
SYS-1244: Process audio files

akohler:
Implements [SYS-1244](https://jira.library.ucla.edu/browse/SYS-1244).

This PR adds full processing of audio files: generating mp3 submasters from master wav files, copying to appropriate directories, and updating the database.  This is done via a new `AudioFileHandler` class, which creates the mp3 and coordinates the relationship between master and submaster.  The file / directory handling and database updates are all done through the `OralHistoryFile` class.

Notable changes:
* New `BaseFileHandler` abstract class, which only exists to enforce `process_files()` existence in subclasses.
* New 'AudioFileHandler` subclass, which implements behavior specific to processing audio files:
  * mp3 generation
  * submaster `OralHistoryFile` creation, as a child of the corresponding master OHF
  * enforce new requirement that audio items can have only one master file
    * There's a bug which allows masters of other types (pdf, image, text) to be attached to audio items.  This will get fixed later, via [SYS-1248](https://jira.library.ucla.edu/browse/SYS-1248).
    * Since the "only one master" rule applies only to audio (right?), `_item_has_files()` is only in `AudioFileHandler`.
* `OralHistoryFile` class now hides data members intended to be read-only by consumers, exposing appropriate access through properties.
* Added `original_file_name` to `MediaFile` model, a required field which stores the original name of the uploaded file.  The same name is stored for masters and derivatives, so we don't lose that info when giving files generic ARK-based names.
  * This added a migration.
* Updated `process_file` management command to use `AudioFileHandler` for audio files.  Other content types are still handled by legacy code which only processes masters; this will be updated as new file handlers are added.
* Tests were updated / added as needed.

Minor changes:
* File upload form now gets `OH_FILE_SOURCE` from settings (set at application start), rather than in real time from environment.
* File upload page now displays file information "better"; more cleanup / improvement will come later.

Testing:
- Restart application to apply migration.
- Test manually as usual.
- Run automatic tests; there should be 37, all passing; 2 tests which previously were skipped are now implemented.


kjallen:
Added additional test for unbound audio submasters and exception catch

kjallen (approved):
Discussed some future considerations for handling bad behavior with ffmpeg, for now the non-duplicate master audio file will always be uploaded regardless of ffmpeg behavior

PR #57, opened 2023-05-03
SYS-1256: Item type cleanup

akohler:
Implements [SYS-1256](https://jira.library.ucla.edu/browse/SYS-1256).

This PR makes improvements to item types:
* Remove unused `Collection` type from the `item-type-data` fixture.
* `Series` is now the top-level item type; otherwise, the hierarchy is unchanged.
* Allow only appropriate item types to display via `ProjectItemForm`, used for adding and editing "pure" item data (not the optional metadata).
  * The only item type allowed for top-level items is `Series`
  * The only item type allowed for 2nd-level items is `Interview`
  * The only item types allowed for 3rd-level items are `Audio` and `Video`, with `Audio` as the default
* Removed the now-obsolete `get_default_type()` method, which always set `Audio` on the form.
* Update existing tests to use real `ItemType` objects instead of id numbers.
* Add tests to confirm the new behavior is correct.
* Minor change: Add `messages` display to `edit_item` template, and a success message, for some visual confirmation that data was saved.
* Bump version to `v0.7.5` for deployment.

Technical details:
`ProjectItemForm.type` is a `ModelChoiceField`, which uses a queryset to determine which `ItemType` values appear in the form.  Instead of all values, this is now dynamic based on where the current item (if any) is in the `Series` -> `Interview` -> `A/V file` hierarchy.

This is determined by passing the current item's parent to the form's `__init()__` method.  Since the item submitted via form may not exist yet, if an item is being created rather than edited, the parent (if any) must be used.  A new method on the form, `_get_relevant_item_types()`, implements this:
* If no parent, choices are `Series`
* If parent is `Series`, choices are `Interview`
* If parent is `Interview`, choices are `Audio` and `Video`

`_get_relevant_item_types()` returns a queryset, which is assigned to the form's `type` field.  The initial value for that field is set to the first value in the queryset, to avoid defaulting to `---------------`.

The form must now have `parent_item` passed to it whenever relevant.  Everywhere this happens in `views` and `views_utils` was updated, except for `get_edit_item_context()` which already has `item.parent` data.

Testing:
Restarting is not needed.  That will load the updated item type fixture, but doesn't remove the obsolete `Collection` type.  That doesn't matter, since `Collection` is completely unused.
```
$ docker-compose exec django python manage.py test
... 43 tests, all passing ...
```

To be thorough, I then deleted `Collection` and re-tested:
```
$ docker-compose exec django python manage.py shell
>>> ItemType.objects.get(type="Collection").delete()
(1, {'oh_staff_ui.ItemType': 1})
```
All 43 tests still pass :)

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v0.7.0
+  tag: v0.7.5
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/fixtures/item-type-data.json
+++ b/oh_staff_ui/fixtures/item-type-data.json
@@ -1,18 +1,10 @@
 [
     {
         "model": "oh_staff_ui.ItemType",
-        "pk": 1,
-        "fields": {
-            "type": "Collection",
-            "parent": null
-        }
-    },
-    {
-        "model": "oh_staff_ui.ItemType",
         "pk": 2,
         "fields": {
             "type": "Series",
-            "parent": 1
+            "parent": null
         }
     },
     {
--- a/oh_staff_ui/forms.py
+++ b/oh_staff_ui/forms.py
@@ -1,6 +1,7 @@
 from django import forms
 from django.conf import settings
 from django.core.cache import cache
+from django.db.models import QuerySet
 from oh_staff_ui.models import (
     AltIdType,
     AltTitleType,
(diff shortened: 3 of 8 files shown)


ztucker4 (approved):
Hi @akohler, this looks good to me. As we discussed, there is some weirdness with the Add Item page that we will resolve later. For now, the Type dropdown is appropriately limiting values based on the parent item's Type. Everything still works after deleting Collection manually, and tests make sense and pass. I also like the new save confirmation on the edit item page. 
Merging, thanks!

PR #59, opened 2023-05-04
SYS-1246: Finish file handling code

akohler:
Implements [SYS-1246](https://jira.library.ucla.edu/browse/SYS-1246) and [SYS-1255](https://jira.library.ucla.edu/browse/SYS-1255).  Bumps version to `v0.8.0` for deployment.

This PR combines related work from two tickets:
1. SYS-1246: Add `GeneralFileHandler` for all files other than audio and image.
2. SYS-1255: Add `MediaFileType.file_code`, which is intended to be a permanent unique human-readable identifier, and use it everywhere `MediaFileType.file_type` was used.  This reduces chances that staff can make changes (like `MasterAudio1` to `Master Audio`, for example) which break code relying on that string as an identifier.  The code now uses values like `audio_master` which will not be exposed to the user interface.

Implementation notes:
* `GeneralFileHandler.create_submaster()` makes a copy of the master file in `/tmp` before adding the corresponding `OralHistoryFile`.  This is not necessary, but it seemed appropriate to do since the other handlers also create files in `/tmp`, so doing it here is consistent and adds minimal overhead.
* Now that all file handlers are written, it may make sense to refactor the similar parts of their `process_files()` methods, pulling similar code into the base handler class.  That's future work.

Minor changes:
* `file_upload` template display of associated files is a little cleaner.
* `file_upload` form now shows only options for "master" files, based on the new codes.
* `manage.py process_files -h` now lists the same master code options for `--file_type`.
* `MediaFileType` admin form hides `file_code` so staff can't accidentally edit it.

Testing:
* Restart application to apply migration for `MediaFileType` changes and to load the updated fixture.
  * The restart should be enough, but if local db is not the same as current `main`, run `docker_scripts/nuke_dev_db.sh`.
* There currently are 48 tests, all passing.
* New tests include basic ones for general files (pdf and text/xml).
* Test manually by confirming that pdf/text/xml files can be uploaded and are processed correctly.


kjallen (approved):
- Removed containers, rebuilt, nuked db for clean environment
- Migrations applied successfully. Ran tests, 48 passed
- Confirmed file_code use
- Tested UI and command line to confirm upload behaves as previously defined, leaf nodes limited to 1 media file and any number of pdf/text file_codes

PR #60, opened 2023-05-04
SYS-1258: Enhance search options

ztucker4:
Implements [SYS-1258](https://jira.library.ucla.edu/browse/SYS-1258)

This PR includes changes to support item search by Status and Keyword in addition to the existing ARK and Title searches. To test, just pull and try new searches - no restart needed.

Changes include:
* Changes to `views.py`and `views_utils.py` to support new search types. Keyword search uses the same algorithm as title keyword search, but checks against all text fields on these models: ProjectItem, AltId, AltTitle, Description, Name, Subject. ProjectItems that use any of the items retrieved in this search are then presented to the user.
* New JS (in `main.js`) to dynamically determine whether the text search box or the Status dropdown is visible in the Search form, depending on the selected "search type".
* New JS (in `validation.js`) to validate the search form and avoid breaking search with empty or all-whitespace searches.
* Template updates to use new validation and remove reference to limited search types in navbar.


akohler (changes requested):
@ztucker4 This is great.  I feared keyword performance would be slow, but it seems just as fast as the other searches.

There are some changes which should be made, mostly cleanup but the main one being de-duping of keyword search results.

I also see cases where no keyword results are found, despite all of the search terms being in indexed fields.  For example, this record http://127.0.0.1:8000/item/2322
contains the three terms `nwonye`, `ashe`, and `fox`in various description fields.  But the search
http://127.0.0.1:8000/search_results/keyword/Nwonye%20fox%20ashe
finds no results.
http://127.0.0.1:8000/search_results/keyword/Nwonye%20fox
does find the expected record.
http://127.0.0.1:8000/search_results/keyword/Nwonye%20ashe
finds about 20 records (correctly, I think), including the expected `fox` record.

ztucker4:
Hi @akohler, this is ready for re-review. Keyword search results are now de-duped, the search form JS now only runs on the search page, and other code cleanup is done. No changes made to the keyword search functionality (for now). Thanks!

akohler (approved):

PR #61, opened 2023-05-04
SYS-1248: Validate content type against file type

akohler:
Implements [SYS-1248](https://jira.library.ucla.edu/browse/SYS-1248).

This PR adds an internal method to `OralHistoryFile` which validates the content type against the file type.  If the check fails, a `ValueError` is raised.

Examples:
* Uploading "file.pdf" but selecting "MasterAudio1" should throw an error
* Uploading "file.wav" and selecting "MasterAudio1" should be accepted.

Testing:
4 new tests were added, checking `audio`, `image`, `pdf`, and `text` (content type) input files against a mismatched file type (e.g., `sample.wav` (audio) against `pdf_master` (pdf).
* No restart necessary.
* `$ docker-compose exec django python manage.py test` should show 52 tests, all passing.
* Pick a random item, like http://127.0.0.1:8000/upload_file/7647, and try to upload a similar mismatch.
* Confirm a matching upload does still work.

### DIFF
--- a/oh_staff_ui/classes/OralHistoryFile.py
+++ b/oh_staff_ui/classes/OralHistoryFile.py
@@ -29,6 +29,8 @@ class OralHistoryFile:
         self._item = ProjectItem.objects.get(pk=self._item_id)
         self._content_type = self.get_content_type(self._original_file_name)
         self._target_dir = self.get_target_dir(self._file_use, self._content_type)
+        # Combination validation checks.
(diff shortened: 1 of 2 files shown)


ztucker4 (approved):
Hi @akohler, this looks good. I tried various combinations of files and file types, and mismatches were correctly rejected. Tests make sense and pass. Merging, thanks!

PR #62, opened 2023-05-04
SYS-1254: Add ability to search / filter admin metadata

ztucker4:
Implements [SYS-1254](https://jira.library.ucla.edu/browse/SYS-1254)

This PR adds a search box (by value) and a filter (by source) to Name and Subject in the admin interface. Combining search and filter values works as you'd expect. Filter values are limited to those actually used in the data. Search by Source is not currently supported, but could be added if desired.


akohler (approved):
@ztucker4 Perfect, works better than expected as it even does substring searches.  Searching by Source is not needed; the filter does that very well.  Merging.

PR #63, opened 2023-05-04
SYS-1262: Capture file size and display to users

akohler:
Implements [SYS-1262](https://jira.library.ucla.edu/browse/SYS-1262).
Bumps version to `v0.9.0` for deployment.

This PR adds `file_size` to the `MediaFile` model, to capture the size in bytes of files as they are processed.  Since files may not always be accessible by the application - master files in particular, which get moved to isolated storage - the application can't rely on `MediaFile.file.size` to retrieve the file size later.

The `file_upload` template was updated to display file size.  Tests which already checked file size on `MediaFile` creation were updated so they now check both the actual file's size at runtime and the captured size.

There's a new migration, since a model was updated.

Testing:
* Restart application to apply migration.
* Run tests: should be 52 tests, all passing.
* Upload a file via the UI and confirm that file size displays, and that the master's reported size matches the actual size of the uploaded file.


oh_staff_ui/templates/oh_staff_ui/upload_file.html
```diff
@@ -44,6 +44,7 @@
     <th>Added by</th>
     <th>Original name</th>
     <th>New name & location</th>
+    <th>Size</th>
```
ztucker4, line 47: Either here or in the table itself, can we say that the the size is given in bytes?

ztucker4 (commented):
Tested uploading a few different files, and sizes are reported correctly. Tests are passing and model changes look good. I have one comment on the template - I'd like to add units to the size display.

oh_staff_ui/templates/oh_staff_ui/upload_file.html
akohler: Good catch.  I've just pushed a change which adds "bytes" to the display.   We might move it to the table header later, when this gets restyled.

ztucker4 (approved):
Looks great now. Merging, thanks!

PR #64, opened 2023-05-05
SYS-1257: Hide metadata fields on file-level items

ztucker4:
Implements [SYS-1257](https://jira.library.ucla.edu/browse/SYS-1257)

This PR hides all optional metadata fields other than Description on Audio/Video ProjectItems. In addition, the only visible qualifier is TableOfContents. Styling is done with CSS, with accompanying changes to the template and JS to add class values. 

One question: should adding multiple Descriptions on file-level items be supported? This is possible for now (though all must be TableOfContents), but we could also hide the plus button.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -69,7 +69,7 @@ td {
     margin-bottom: 3px;
  }
 
-.file_metadata tr:not([id="description_row"]) {
+.file_metadata tr:not([class="description"]) {
     display: none;
 }
 
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -69,11 +69,15 @@ td {
     margin-bottom: 3px;
  }
 
+/* .file_metadata: optional metadata on edit_item page
+for Audio and Video items. Hide all table rows other 
+than Description. */ 
 .file_metadata tr:not([class="description"]) {
     display: none;
 }
-
-.file_metadata select option:not([value="3"]){
+/* for file-level metadata, hide all dropdown options
+other than TableOfContents (value = "3") */ 
+.file_metadata .description select option:not([value="3"]){
     display: none;
 }
 
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -69,6 +69,14 @@ td {
     margin-bottom: 3px;
  }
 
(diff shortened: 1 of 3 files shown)


static/css/style.css
```diff
@@ -69,6 +69,14 @@ td {
     margin-bottom: 3px;
  }
 
+.file_metadata tr:not([class="description"]) {
+    display: none;
+}
+
+.file_metadata select option:not([value="3"]){
```
akohler, line 76: Please add a comment here, to explain that "3" represents table of contents.  Using the numeric value is less fragile than using the display value "tableOfContents", but needs explanation.

static/js/main.js
```diff
@@ -30,6 +30,7 @@ function showEmptyForm(event) {
     // Remove id & class, copied from emptyForm
     newForm.removeAttribute("id");
     newForm.classList.remove("empty_form");
+    newForm.classList.add(metadataType);
```
akohler, line 33: I'm not seeing where this added class is being used, when I inspect or view source it doesn't appear, except for the explicitly added `class = "description">` from line 197 in `edit_item.html`.  Can you please clarify how this is used?  It also conflicts with the comment & lines of code above it, which remove things, not add them.

Not necessarily wrong, but needs clarification if nothing else.

akohler (changes requested):
@ztucker4 Functionally, this all seems to work: the display is limited correctly, I can still add / edit / delete the description field on file-level and higher level items, and editing other metadata still works as expected.

I'm asking for more comments in code to clarify some things, then I'll merge.

Thanks --Andy

static/js/main.js
ztucker4: `class = "description"` matters due to the first new CSS rule, which hides all `<tr>`s of optional metadata on file-level items unless they have that class. I've added a comment addressing this and separating it from the removals above.

ztucker4:
Hi @akohler, this is ready for another look. I've added comments to the CSS and JS involved in this change. Please let me know if anything still isn't clear. Thanks!

akohler (approved):

akohler:
@ztucker4 Thanks for clarifying; merged.

PR #65, opened 2023-05-05
Initial add of MODS export libraries and tests

kjallen:
Very basic PR to confirm install and base behavior

- [eulxml](https://eulxml.readthedocs.io/) added to requirements.txt
- `OralHistoryMods` class added extending base `eulxml.xmlmap.mods.MODS` class
- Tests added to validate empty MODS records pass for each item type

This PR is primarily to confirm libraries are installed and able to be tested properly.

To confirm run:
`docker-compose exec django python manage.py test`

All tests should pass if the `eulxml` library is properly installed and a valid empty MODS record can be created for each item type. Jira ticket will be updated to reflect the stripped down PR.

For now the version of eulxml is `1.1.3` rather than the most recent github commit, to be revised if required.
ply version `3.11` added to resolve warning.

### DIFF
--- a/oh_staff_ui/tests.py
+++ b/oh_staff_ui/tests.py
@@ -737,6 +737,48 @@ class ProjectItemFormTestCase(TestCase):
         # Check that form validation succeeded.
         self.assertEquals({}, form.errors)
 
+
+class ModsTestCase(TestCase):
+    # MODS related tests require an item of each type to confirm proper record creation,
+    # each has a slightly different field structure in the MODS record
+
+    # Load the lookup tables needed for these tests.
+    fixtures = [
+        "item-status-data.json",
+        "item-type-data.json",
+    ]
+
+    @classmethod
+    def setUpTestData(cls):
+        # Use QAD data for fake user and fake items.
+        cls.user = User.objects.create_user("tester")
+        # Level 1: Series.
+        cls.series_item = ProjectItem.objects.create(
+            ark="fake/abcdef",
+            created_by=cls.user,
+            last_modified_by=cls.user,
+            title="Fake series",
+            type=ItemType.objects.get(type="Series"),
+        )
+        # Level 2: Interview, child of series.
+        cls.interview_item = ProjectItem.objects.create(
+            ark="fake/abcdef",
+            created_by=cls.user,
+            last_modified_by=cls.user,
+            title="Fake interview",
+            type=ItemType.objects.get(type="Interview"),
+            parent=cls.series_item,
+        )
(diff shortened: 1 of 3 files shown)


oh_staff_ui/tests.py
```diff
@@ -734,3 +736,24 @@ def test_post_to_item_form(self):
         form = ProjectItemForm(request.POST, parent_item=self.series_item)
         # Check that form validation succeeded.
         self.assertEquals({}, form.errors)
+
+    def test_valid_series_item_mods(self):
```
akohler, line 740: These new tests should go into a new test case, since they don't really correspond to `ProjectItemFormTestCase`.  Make a new test case and copy `setUpTestData` , then move these tests to that case.

requirements.txt
```diff
@@ -5,3 +5,4 @@ whitenoise == 6.0.0
 requests == 2.28.2
 pillow == 9.5.0
 ffmpeg-python == 0.2.0
+eulxml == 1.1.3
```
akohler, line 8: I see this is quite old - June 2016, though they made some changes in Dec 2021 for "Django 3 compatibility".  If it works for us, OK - but I see a lot of forks and old open issues and PRs, so there may be little / no support.

Tests pass, though I get a warning `WARNING: yacc table file version is out of date`.

akohler (changes requested):
@kjallen Please move the tests into a new test case just for MODS work.

requirements.txt
kjallen: Thanks, i'll update tests and confirm version -- even with the github comments, the MODS schema is ahead of what is currently being used now (mods-3.3 vs mods-3.4), it does offer many convenience wrappers for OAI and MODS, noted. I did not receive the `yacc` warning, i'll try to reproduce after pulling into new test case.

kjallen: For now, using eulxml 1.1.3, but can pull from most[ recent commit ](https://github.com/emory-libraries/eulxml/tree/2e1a) if preferred.
I've set the ply library to a specific version, this should resolve yacc warning.

oh_staff_ui/tests.py
kjallen: `ModsTestCase` class created and cases moved

requirements.txt
akohler: Rebuilding after the `ply` explicit reference, the `yacc` warning is gone.
Let's stick with the official release of `eulxml`; if we run into problems with more extensive use, we can consider pulling from github instead of pypi, or more extreme measures.

akohler (approved):

PR #66, opened 2023-05-05
SYS-1273: 

ztucker4:
Implements [SYS-1273](https://jira.library.ucla.edu/browse/SYS-1273)

Instead of checking that the current page's URL contains "item_search" (which missed the home page search form), `main.js` now checks for the existence of the item_search form before executing the code specific to this page. This fixes the duplicated "Query:" label, and Status search now works directly from the homepage.


akohler (approved):
Fix confirmed, merging.

PR #69, opened 2023-05-10
Adding ark as identifier to MODS record

kjallen:
Basic PR to rework some tests and add an ARK as an identifier field in a MODS record

Some very slight changes to OralHistoryMods class
- Constructor accepts `ProjectItem` record rather than an id
- `populate_fields` public function intended to house each function to populate a field
- `_populate_identifier` private function that other MODS fields will follow a similar pattern
- Tests updated to reflect constructor changes

This PR is basic to get template ironed out, the remainder of fields will follow a similar pattern, a private function that performs any logic required to populate that particular MODS field, then added to the `populate_fields` function with related tests.

### DIFF
--- a/oh_staff_ui/classes/OralHistoryMods.py
+++ b/oh_staff_ui/classes/OralHistoryMods.py
@@ -1,11 +1,19 @@
+from eulxml.xmlmap import mods
 from eulxml.xmlmap.mods import MODS
-
+from oh_staff_ui.models import ProjectItem
 import logging
 
 logger = logging.getLogger(__name__)
 
 
 class OralHistoryMods(MODS):
-    def __init__(self, ark):
+    def __init__(self, project_item):
         super().__init__()
-        self._ark = ark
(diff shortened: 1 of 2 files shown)


oh_staff_ui/classes/OralHistoryMods.py
```diff
@@ -1,11 +1,19 @@
...
-
+from oh_staff_ui.models import ProjectItem
 import logging
 
 logger = logging.getLogger(__name__)
 
 
 class OralHistoryMods(MODS):
-    def __init__(self, ark):
+    def __init__(self, project_item):
```
akohler, line 10: Please use type hints (`project_item: ProjectItem`).  Not requiring it for this PR but should be use in later ones.
`'oh_staff_ui.models.ProjectItem' imported but unused Flake8(F401) [Line 3, Col 1]`

akohler (approved):
This is fine as a foundation; merging.

PR #70, opened 2023-05-10
SYS-1263: Hide Parent drop-down on add_item form page

ztucker4:
Implements [SYS-1263](https://jira.library.ucla.edu/browse/SYS-1263)

Adds a small amount of conditional JS to hide Parent dropdown and label on the base `add_item` page (i.e. for new series creation). Dropdown and label are still visible after following link to add a child item, and on `edit_item` pages. Base template is also updated to change `add_item` link to read "Add Series".


akohler (approved):
@ztucker4 This looks good, works as expected - merging.

PR #72, opened 2023-05-11
SYS-1296: Style with Bootstrap 5

ztucker4:
Implements [SYS-1296](https://jira.library.ucla.edu/browse/SYS-1269)

Applies Bootstrap styling to the OH Staff UI application using `django-bootstrap5`. To test, restart your Django container to install the new package requirement, then marvel at the new post-Y2K-looking site. Please test both appearance and functionality, as many of the changes interact with our JS.

The styling is largely default Bootstrap, so there is room for improvement. I have a couple of questions here:
* Form fields are larger and more spaced out now. This is probably easier to read, but also means more scrolling for long `edit_item` pages. Should we try reducing the vertical space between form fields?
* Blank text fields now contain placeholder text. I can't tell if the default color of the text is dark enough to potentially cause confusion regarding which fields are filled out. Should we change this to a different color, or remove the placeholders entirely?

### DIFF
--- a/oh_staff_ui/templates/oh_staff_ui/add_item.html
+++ b/oh_staff_ui/templates/oh_staff_ui/add_item.html
@@ -11,10 +11,13 @@
 
 <form name="add_item" id="add_item" method="POST">
     {% csrf_token %}
-    <table>
-        {% bootstrap_form form layout="horizontal" %}
-    </table>
-    <br>
+    {% bootstrap_field form.parent layout="horizontal" %}
+    {% bootstrap_field form.title layout="horizontal" placeholder="" %}
+    {% bootstrap_field form.type layout="horizontal" %}
+    {% bootstrap_field form.sequence layout="horizontal" placeholder="" %}
+    {% bootstrap_field form.coverage layout="horizontal" placeholder="" %}
+    {% bootstrap_field form.relation layout="horizontal" placeholder="" %}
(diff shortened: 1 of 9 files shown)


akohler (changes requested):
@ztucker4 Overall, if people think this looks better, I defer to that... personally, I'm unsure.  Technically, this seems fine, and functionality is all still there.

> Should we try reducing the vertical space between form fields?

Yes, I think this is too spread out.  I commented on a few specific forms, where I think the "horizontal" bootstrap approach is not good.  I agree that for larger forms which do need to be multi-line, the vertical space seems excessive.

> Should we change this to a different color, or remove the placeholders entirely?

I vote for removing the placeholder text.  Since we have labels which convey what the field is for, largely repeating that as a placeholder is redundant and I agree can look like there's real text in the fields.

ztucker4:
Hi @akohler, this is ready for re-review. I've changed the search and file upload forms to display fields horizontally, removed placeholders in text input fields, increased the width of `qualifier`s in the `edit_item` form, and reduced vertical space between form fields (from `1rem` to `.4rem`, so we can reduce further if needed). Thanks!

akohler (approved):

akohler:
@ztucker4 The changes look great.  The smaller forms are much cleaner, and the reduced space between multi-line form elements helps.  Removing the placeholder text from the item editing form makes empty fields much clearer.  Merging.

PR #73, opened 2023-05-12
SYS-1281: Add URLs for non-master file access

akohler:
Implements [SYS-1281](https://jira.library.ucla.edu/browse/SYS-1281).
Bumps version to `v0.9.2` for deployment.

This PR adds a new property on the `MediaFile` model, `file_url`.  This is calculated based on the full path to the file (relative to the container), which is available via `MediaFile.file.name`.  The URL is not stored in the database, and cannot be queried - it's only available via an instance of `MediaFile`.

`MediaFile.file_url` currently is determined only for non-masters, pending policy and technical decisions about future access to master files.  It also has meaning only for production files (whether in production environment, or data loaded in dev environment), since all real files are accessed by separate production web servers.

There's also a new property, `MediaFile.file_name_only`, which provides the bare file name with no path, for display in templates (e.g., `file_name.xml`, not `/long/container-specific/path/to/file_name.xml`).

The `upload_file` template now displays `file_name_only`, hyperlinked if `file_url` has a value.

There are 4 new tests, covering the 3 valid cases (submaster wowza audio, submaster static files (including images), thumbnail images) and 1 invalid / unsupported case (master files).  The tests are a little awkward, as I found that full paths outside of Django's `BASE_DIR` can raise a `SuspiciousFileOperation`... which I'm still investigating, via SYS-1286.

Testing:
There are 62 tests, all passing.
If you have production file data loaded (via a recent nuke rebuild, or via `docker-compose exec django python manage.py import_file_metadata migration_data/file_metadata.tsv`, look at a few items with files and verify the URLs work as expected.
Examples:
* http://127.0.0.1:8000/upload_file/2339 (PDF and images)
* http://127.0.0.1:8000/upload_file/9850 (audio and XML)

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v0.9.1
+  tag: v0.9.2
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/models.py
+++ b/oh_staff_ui/models.py
@@ -536,3 +536,35 @@ class MediaFile(models.Model):
             if Path(new_name).exists() or MediaFile.objects.filter(file=new_name):
                 raise FileExistsError(f"File already exists: {new_name}")
         super().save(*args, **kwargs)
+
+    @property
+    def file_name_only(self) -> str:
+        # Convenience method which returns just the file name, not including the full path.
+        # Used for display in templates.
+        return Path(self.file.name).name
+
+    @property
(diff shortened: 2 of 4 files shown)


kjallen (approved):
- Wiped and rebuild DB
- Reviewed commits
- Ran tests, all passed
- Uploaded sample item of each type in UI, verified shortened file name column and links
- Verified new field on several items in shell:
```
>>> from oh_staff_ui.models import MediaFile
>>> mf = MediaFile.objects.get(id=555)
>>> mf.file_name_only
'21198-zz0008zdcp-2-submaster.xml'
>>> 
```

PR #75, opened 2023-05-15
SYS-1286: Fix file handling

akohler:
Implements [SYS-1286](https://jira.library.ucla.edu/browse/SYS-1286). Bumps version to `v0.9.3` for deployment.

This PR fixes a major bug which prevented file uploads from working in the production environment.  The bug was caused by my misunderstanding of Django's `FileField` and its `upload_to` functionality, along with a significant difference in where uploaded files go, between development and production environments.

I thought that `MediaFile.file(upload_to="/some/absolute/path/to/file")` would (a) work, and (b) mean that Django's `MEDIA_ROOT` didn't need to be defined.  Both were wrong :(

`MEDIA_ROOT` defaults to `''`, meaning the top-level directory in the project.  In the development environment, files were going to `media_dev/` - in the top-level directory - with relative paths like `media_dev/oh_static/etc.`.  This all worked fine.

In production, files need to go into `/media/some/path/etc`, which is outside the Django project.  With the default `MEDIA_ROOT`, this is not allowed, causing a `SuspiciousFileOperation` exception.  And `MediaFile.file(upload_to=` does not accept absolute paths, just relative ones.

So, finally, this PR:
* Sets `MEDIA_ROOT` via environmental variable
  * In production: `/media`
  * In development: `/tmp/media_dev` (outside the project, to better simulate the prod environment)
* Updates related path variables like `OH_STATIC` to be relative, not absolute
* Removes the `get_target_path()` method which `MediaFile.file(upload_to=` was (mis)using
  * This required updating a past migration file to make Django happy; no actual migration is needed
* Removes / simplifies related code which no longer needs to distinguish between dev and prod paths
* Updates all relevant tests
* Removes `OH_LIBPARTNERS` and `USE_TEST_DIRS` variables, not needed in this project
* Removes the local `media_dev` directories, no longer needed
* Updates `README.md` with relevant info on viewing `/tmp/media_dev`, since that's only in the container

Testing:
* Pull and check out this branch, `SYS-1286/fix_upload_error`
* Run `docker_scripts/nuke_dev_db.sh` to clear old data, including incorrect file metadata
* Run tests: `docker-compose exec django python manage.py test`; should be 62 tests, all passing
* Upload a few sample files to confirm that all still works correctly via UI
* Check a few existing items and their files, like http://127.0.0.1:8000/upload_file/823 , to confirm URLs for migrated derivative metadata are correct

### DIFF
--- a/.docker-compose_django.env
+++ b/.docker-compose_django.env
@@ -43,5 +43,3 @@ DJANGO_OH_MASTERSLZ=oh_lz
 DJANGO_OH_WOWZA=oh_wowza
 # Static files
 DJANGO_OH_STATIC=oh_static
-# Oral History source files (not used in dev environment)
-DJANGO_OH_LIBPARTNERS=oh_source
--- a/.docker-compose_django.env
+++ b/.docker-compose_django.env
@@ -28,28 +28,20 @@ DJANGO_SUPERUSER_EMAIL=softwaredev-systems@library.ucla.edu
 # URL to ARK minter (PROD)
 DJANGO_ARK_MINTER=http://noid.library.ucla.edu/noidu_zz8+?mint+1
 
-# Directory containing files "uploaded" by staff.
+# Directory containing files to be "uploaded" by staff.
 # Relative path, starting top level of Django project
 DJANGO_OH_FILE_SOURCE=samples
 # Or absolute path, if preferred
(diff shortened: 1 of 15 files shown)


ztucker4 (approved):
Hi @akohler, this looks good.
* Updated tests pass
* File upload works correctly through the UI - files appear in /tmp/media_dev as expected
* URLs for migrated derivatives are correct
Merging, thanks!

PR #77, opened 2023-05-16
SYS-1276: Migrate original DLCS user names as-is

akohler:
Implements [SYS-1276](https://jira.library.ucla.edu/browse/SYS-1276).

This PR changes how DLCS user information is migrated.  Originally, DLCS item creation and modification user values were mapped to email addresses for valid DLCS user records, with unmappable values mapped to a generic `oralhistory` user.  This loss of historical user data was unacceptable by stakeholders.

With this change, the original `PROJECT_ITEM.CREATED_BY` and `PROJECT_ITEM.MODIFIED_BY` values are preserved and migrated as-is to create Django user names.  Post-migration, some data cleanup may be done in the new system to make these users more consistent.

Changes:
* Updated export script `export_projectitems.py` query to pull the above values as-is instead of mapping to email addresses.
  * Also reformatted some unrelated parts of the query for better readability.
* Updated import management command `import_projectitems.py` to create user records from these raw values instead of from email addresses.
* Exported new data file `migration_data/project-items-export.tsv` and used it to test & confirm the changes.
* Exported new fixture `user-data.json` with the newly-created users.  This lets us load user data without having to import all of the item data.
  * The fixture also sets all of these users to inactive, as opposed to the default Django default of active.
  * Login is impossible for these user names without manual intervention by an administrator.
  * Passwords for these users are set to an invalid unhashed value as well.
 
Testing:
Reviewing the query and code change should be enough, but if desired:
```
# No migrations, but full data reload is needed
$ docker_scripts/nuke_dev_db.sh
```
All 62 tests should still pass (no code covered by tests was changed).
View an item like http://127.0.0.1:8000/item/815 to confirm that it was created by `oralhistory data entry` instead of `oralhistory`.
Check other items if desired.


kjallen (approved):
- Reviewed query
- Nuked local DB
- Confirmed several items for proper created by username

PR #80, opened 2023-05-18
Addition of alternate title mods element

kjallen:
Addition of AltTitle mods element.

If an AltTitle record is present an additional `mods:titleInfo` field will be created with type of `alternative`.

Testing:

`AltTitle` is not as common as other fields, so a quick way is to edit a record to add an AltTitle with any qualifier, and confirm via shell using similar calls as previous PR.

```
>>> pi = ProjectItem.objects.get(id=2207)
>>> ohm = OralHistoryMods(pi)
>>> ohm.populate_fields()
>>> ohm.serializeDocument()
```
A test has been added to check for the presence of `titleInfo` along with an attribute type `alternative`, different from the standard `title` field.

### DIFF
--- a/oh_staff_ui/classes/OralHistoryMods.py
+++ b/oh_staff_ui/classes/OralHistoryMods.py
@@ -2,6 +2,7 @@ import logging
 from eulxml.xmlmap import mods
 from eulxml.xmlmap.mods import MODS
 from oh_staff_ui.models import (
+    AltTitle,
     Date,
     Description,
     ItemLanguageUsage,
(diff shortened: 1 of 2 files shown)


akohler (approved):
Looks good to me.  Thanks for refactoring the tests.  Also tested manually with item 306, which has an alternate title, and saw expected output:
```
 <mods:titleInfo type="alternative">
    <mods:title>California State Archives State Government Oral History Program: Interview with Frank Cullen Sr.</mods:title>
  </mods:titleInfo>
```
Merging.

PR #81, opened 2023-05-19
Addition of test and alt identifier mods element

kjallen:
Addition of AltId mods element.

Testing:

Either add alternate id or use an existing record with one, there are a handful with `OPAC` types
Interview of Cecil Talney - `id=1482`, this may not be same in local DB.
```
>>> pi = ProjectItem.objects.get(id=1482)
>>> ohm = OralHistoryMods(pi)
>>> ohm.populate_fields()
>>> ohm.serializeDocument()
```
A test has been added to check for the presence of identifier along with an assigned attribute


akohler (approved):
@kjallen Tests and output confirmed, using item 1482:
``<mods:identifier type="OPAC">4433051</mods:identifier>```

PR #86, opened 2023-05-23
SYS-1270: Process audio files in background

akohler:
Implements [SYS-1270](https://jira.library.ucla.edu/browse/SYS-1270).  Bumps version to `v0.9.6` for deployment.

Audio WAV master files are usually several hundred megabytes in size, and full processing can take a few minutes.  This PR moves that processing to a separate thread which runs in the background, returning immediate control of the browser to the user.

### Major changes:
* Audio file processing now runs via separate thread.  Other file processing is unchanged.
* The `process_file` management command is now called via `views_utils.run_process_file_command()`, providing a callable needed for running via a thread.  This wrapper also explicitly closes the db connection used, which can otherwise lead to problems when running via thread.
* Added a new `MediaFileError` model to capture file processing errors to database, allowing them to be displayed to the user at any time.  This currently is used only for audio processing errors.
  * Added test to confirm this error message is captured.
* Updated `upload_file` template to display these errors, when present for an item.
* Moved `upload_file` form validation to the form's `clean()` method, ensuring all validation-related errors (content type mismatch, unknown media type) are caught and displayed immediately.

### Smaller changes:
* Added `file_size` property to `OralHistoryFile` class, along with tests and minor related refactoring.
* Cleaned up / improved various comments in related code.

### Testing:
Restart application to apply `MediaFileError` migration.
Run tests - should be 76, all passing.

Try uploading sample files incorrectly to confirm those error messages still display immediately.  For example:
* `bad_extension.foo` as `MasterAudio1` - unknown content type
* `sample.pdf` as `MasterAudio1` - file/content type mismatch

Next, try uploading one of the "fake" audio files, like `fake_audio_02.wav`.  A message should immediately appear:
`samples/fake_file_02.wav is being processed in the background. You can continue editing while this is happening.`
Refresh the page to see that
* The master was added (unfortunate; maybe we'll add WAV content validation someday)
* A `File Errors` table appears with detailed error message, similar to this screenshot:
oh_audio_failure
 
Finally try uploading a large WAV file.  @kjallen , I copied one of the WAV files currently in the real OH uploads folder.  Again, you should get an immediate "processing in the background" message.  After 1-2 minutes (depending on your local machine), refresh / revisit the page and you should see successful processing results, similar to this screenshot:
oh_audio_success


kjallen (commented):
Minor revision discussion via slack - follow same message display paradigm for multiple wav upload attempt as corrupt wav/mp3, listing as comment for now rather than request for change.

Review/test steps:
- Restarted app, confirmed migration applied
- Ran tests, all passed
- Confirmed 'bad' wav file proper behavior
- Confirmed proper behavior for valid files
- Confirmed navigation/editing, message and file records on large (> 1GB file)
- Reviewed code

kjallen (approved):
Confirmed merge conflicts resolved

PR #87, opened 2023-05-24
SYS-1290: Add parent title to search results

ztucker4:
Implements [SYS-1290](https://jira.library.ucla.edu/browse/SYS-1290)

Small view, template, and CSS changes to support display of item Parent in search results. Some notes:

* Parents are not currently linked, in keeping with how Title is displayed.
* If an item has no parent, the Parent column is empty.
* Using Bootstrap table styling was easier than dealing with potentially very long lines with text wrapping, so I've applied that to the search results table and removed two now-unused CSS rules.

### DIFF
--- a/oh_staff_ui/templates/oh_staff_ui/search_results.html
+++ b/oh_staff_ui/templates/oh_staff_ui/search_results.html
@@ -14,7 +14,7 @@
     <tr>
         <td><a href="{% url 'edit_item' item.id %}">{{ item.ark }}</a></td>
         <td>{{ item.title }}</td>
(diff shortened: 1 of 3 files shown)


akohler (approved):
@ztucker4 Looks fine to me.  Leaving Parent unlinked is consistent; we'll see if the users request changes.  I confirmed that the removed `search-results` CSS rules aren't used elsewhere.  It would be nice not to have the data/logic difference for `parent` vs. `parent__title`... but a quick look at the keyword search reminded me why that's not worth worrying about....

And thanks for the comments explaining the context for the differences!

Merging --Andy

PR #89, opened 2023-05-24
SYS-1291: Provide ability to browse collection

ztucker4:
Implements [SYS-1291](https://jira.library.ucla.edu/browse/SYS-1291)

This PR adds a new page, accessible through a new "Browse Series" link in the navbar, with a tree containing links to all Series and Interviews (but not Files). This new page reuses the `item_tree` template and includes a bit of text at the top intended to help set user expectations - let me know if you think this should be edited/better formatted/removed.

Screenshot:
image


akohler (approved):
@ztucker4 This is great.  Wording is fine with me.  Performance and display all work as expected.  The code and comments are clear.  There's some conceptual overlap with `get_descendants()`, but enough difference that I think it's not worth pursuing.  Merging!

PR #90, opened 2023-05-24
SYS-1296: Change resource label on item editing form

akohler:
Implements [SYS-1296](https://jira.library.ucla.edu/browse/SYS-1296).
Bumps version to `v0.9.7` to deploy this and other recent UI changes.

This PR changes the label from `Resource(s)` to `Type(s) of Resource` in the `edit_item` template.  It also tweaks the CSS to prevent the longer label text from wrapping.

Testing: No automated tests.  Look at http://127.0.0.1:8000/item/251 or any other item to see the new wording and confirm the label text does not wrap.

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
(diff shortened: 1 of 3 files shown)


ztucker4 (approved):
Looks good - new label is correct and non-wrapping. Merging, thanks!

PR #92, opened 2023-05-25
SYS-1292: Show item type in search results

akohler:
Implements [SYS-1292](https://jira.library.ucla.edu/browse/SYS-1292).  Bumps version to `v0.9.8` for deployment.

This PR adds item type (currently Series, Interview, Audio) to the search results, to make the context of each result clearer.

The original code handled keyword search results differently from other search types, with the `search_results` view and template both needing to account for this difference.  As part of this PR, the logic moves from the view & template to a new utility function, `views_utils.get_search_results()`  This returns consistent data, an appropriately-sorted list of items, so the view can just pass that data to the template for uniform display.

Testing:
No new (or old) automated tests; all tests still pass.
Do searches of different types to confirm `Type` now displays between `ARK` and `Title`.
http://127.0.0.1:8000/item_search/

Screenshot example:
oh_search_results_with_type

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v0.9.7
+  tag: v0.9.8
   pullPolicy: Always
 
 nameOverride: ""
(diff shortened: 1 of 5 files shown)


kjallen (approved):
- Confirmed tests still pass
- Confirmed `Type` does indeed appear in search summary results
- Reviewed code updates

PR #93, opened 2023-05-30
Sys 1211/management cmd mods export

kjallen:
This adds a management command and test and write a single record to a public mods directory specified by `ProjectItem.id`. A future PR on this same ticket will add management options and test for bulk export, calling similar functions.

The exported file name is the item ark with `/` swapped for `-` and `-mods.xml` appended. 
A full filename would look like:
`21198-zz00123456-mods.xml`

This is written to the static root directory under a `mods` subdirectory.
`{MEDIA_ROOT}/{OH_STATIC}/mods/{filename}.xml`

If a file exists, it is overwritten without warning.

To confirm proper behavior:
- Run tests, all should pass. Additional test writes a sample object mods record and checks for existence, calling the write function directly, as opposed to the management command
- Run the `create_mods_records` management command, choose any item id and run the following from the cli, 2504 is used below
`docker-compose exec django python manage.py create_mods_records -i 2504`

A file will be created in the container under `/tmp/media_dev/oh_static/mods` with the pattern described above (`21198-zz002knx8t-mods.xml` in the example command above/

### DIFF
--- a/oh_staff_ui/management/commands/create_mods_records.py
+++ b/oh_staff_ui/management/commands/create_mods_records.py
@@ -1,20 +1,20 @@
 import logging
 from django.core.management.base import BaseCommand
+from django.core.management.base import CommandError
 from oh_staff_ui.classes.OralHistoryMods import OralHistoryMods
 from oh_staff_ui.models import ProjectItem
-
-# For handling command-line processing
-from django.contrib.auth.models import User
-
 
 logger = logging.getLogger(__name__)
 
 
 def create_mods_records(item_id: int) -> None:
     """Given an item_id write the related MODS record to public location"""
-    pi = ProjectItem.objects.filter(id=item_id).first()
-    if pi:
+    try:
+        pi = ProjectItem.objects.get(id=item_id)
         OralHistoryMods(pi).write_mods_record()
+    except (ProjectItem.DoesNotExist) as e:
+        logger.error(e)
+        raise CommandError(f"ProjectItem with id {item_id} does not exist")
 
 
 class Command(BaseCommand):
--- a/oh_staff_ui/management/commands/create_mods_records.py
+++ b/oh_staff_ui/management/commands/create_mods_records.py
@@ -28,8 +28,8 @@ class Command(BaseCommand):
             required=True,
             help="The id of the item to export mods record of",
         )
-    
+
     def handle(self, *args, **options):
-        
+
         item_id = options["item_id"]
         create_mods_records(item_id)
--- a/oh_staff_ui/management/commands/create_mods_records.py
+++ b/oh_staff_ui/management/commands/create_mods_records.py
@@ -1,0 +1,28 @@
+import logging
+from django.core.management.base import BaseCommand
+from oh_staff_ui.classes.OralHistoryMods import OralHistoryMods
+from oh_staff_ui.models import ProjectItem
+
+# For handling command-line processing
+from django.contrib.auth.models import User
+
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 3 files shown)


oh_staff_ui/management/commands/create_mods_records.py
```diff
@@ -0,0 +1,35 @@
+import logging
+from django.core.management.base import BaseCommand
+from oh_staff_ui.classes.OralHistoryMods import OralHistoryMods
+from oh_staff_ui.models import ProjectItem
+
+# For handling command-line processing
+from django.contrib.auth.models import User
```
akohler, line 7: @kjallen This import is not used.  Do you expect to do so in the bulk export?

oh_staff_ui/management/commands/create_mods_records.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def create_mods_records(item_id: int) -> None:
+    """Given an item_id write the related MODS record to public location"""
+    pi = ProjectItem.objects.filter(id=item_id).first()
```
akohler, line 15: `id` is always unique.  The `filter().first()` idiom avoids exceptions, but the user has no info whether this operation succeeded or failed.

It's better to use `get()`, and catch the `ProjectItem.DoesNotExist` exception with a useful message to the user.

oh_staff_ui/management/commands/create_mods_records.py
```diff
@@ -0,0 +1,35 @@
+import logging
+from django.core.management.base import BaseCommand
+from oh_staff_ui.classes.OralHistoryMods import OralHistoryMods
+from oh_staff_ui.models import ProjectItem
+
+# For handling command-line processing
+from django.contrib.auth.models import User
+
+
+logger = logging.getLogger(__name__)
```
akohler, line 10: `logger` is not used.

akohler (changes requested):
@kjallen This works well.  Confirmed tests pass, sample file written look fine to me.

In addition to the minor changes requested, should there be logic in `write_mods_record()` to only write records which qualify, based on status?  Or will that logic be in the management command when the bulk export functionality is added?  Having it in the `OralHistoryMods` class seems best, perhaps via some sort of "OK to publish" flag.

oh_staff_ui/management/commands/create_mods_records.py
kjallen: I did not need it in this commit but will be using to check status in the next PR, I can remove for now.

akohler: It's fine to leave it in if it'll be used later.

oh_staff_ui/management/commands/create_mods_records.py
kjallen: Noted, i'll remove for now, will be using in subsequent PR.

oh_staff_ui/management/commands/create_mods_records.py
kjallen: Swapped `filter().first()` for `get()`, added exception handling and user error message for `ProjectItem.DoesNotExist` case

kjallen:
> should there be logic in `write_mods_record()` to only write records which qualify, based on status? Or will that logic be in the management command when the bulk export functionality is added? Having it in the `OralHistoryMods` class seems best, perhaps via some sort of "OK to publish" flag.

Yes, i'm adding that as part the bulk command now, my initial thinking was if you *want* for force a MODS export on a single item, you can, but will not have option when selecting a bulk export.

akohler (approved):

PR #94, opened 2023-05-31
Reworked mods export management command, added bulk options

kjallen:
This adds PR bulk export MODS option via django management command.

A user may export single items with no restrictions via the `--item` option, or may use `--bulk` with a select number of item types to export MODS record in bulk. 

The bulk options include:
- series
- interview
- audio
- all

If there is a problem writing a record in a bulk setting, the `CommandError` is captured, the `ProjectItem.id` collected, and the remainder of the `ProjectItem` list is exported. The user is shown a result message at the end of operation.

Usage and testing:

The MODS writing logic has not changed, so an additional test has not been added.

To test the management command, run the following:
`docker-compose exec django python manage.py create_mods_records -b series`

Assuming a full test database, the output should be similar to:
`99 of 99 MODS records written`

For a larger set, run:
`docker-compose exec django python manage.py create_mods_records -b all`
`4700 of 4700 MODS records written`

This should only take about 2 minutes.

The content can be verified by looking at the `/tmp/media_dev/oh_static/mods` directory on the django container.

### DIFF
--- a/oh_staff_ui/management/commands/create_mods_records.py
+++ b/oh_staff_ui/management/commands/create_mods_records.py
@@ -17,7 +17,7 @@ class Command(BaseCommand):
             nargs="+",
             type=int,
             required=False,
-            help="The id of the item to export mods record of, multiple ids are permitted",
+            help="The id(s) of the item to export mods record of",
         )
         parser.add_argument(
             "-b",
@@ -43,6 +43,7 @@ class Command(BaseCommand):
             pi = ProjectItem.objects.get(id=item_id)
             OralHistoryMods(pi).write_mods_record()
             logger.info(f"Item: {pi.id} with ark {pi.ark} MODS record written")
+
         except (ProjectItem.DoesNotExist) as e:
             logger.error(e)
             raise CommandError(f"ProjectItem with id {item_id} does not exist")
--- a/oh_staff_ui/management/commands/create_mods_records.py
+++ b/oh_staff_ui/management/commands/create_mods_records.py
@@ -7,16 +7,6 @@ from oh_staff_ui.models import ProjectItem
 logger = logging.getLogger(__name__)
 
 
-def create_mods_records(item_id: int) -> None:
-    """Given an item_id write the related MODS record to public location"""
-    try:
-        pi = ProjectItem.objects.get(id=item_id)
-        OralHistoryMods(pi).write_mods_record()
-    except (ProjectItem.DoesNotExist) as e:
-        logger.error(e)
-        raise CommandError(f"ProjectItem with id {item_id} does not exist")
-
-
 class Command(BaseCommand):
     help = "Django management command to generate Oral History MODS records"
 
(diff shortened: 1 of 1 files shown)


oh_staff_ui/management/commands/create_mods_records.py
```diff
@@ -7,29 +7,63 @@
...
-
-
 class Command(BaseCommand):
     help = "Django management command to generate Oral History MODS records"
 
     def add_arguments(self, parser):
         parser.add_argument(
             "-i",
             "--item_id",
+            nargs="+",
```
akohler, line 17: Since multiple items can be specified, like `-i 1 3 243`, the help should make that clear.  Or remove that capability, since exporting multiples probably should also support the counting and error handling.  I vote for removing it, but up to you.

oh_staff_ui/management/commands/create_mods_records.py
```diff
@@ -7,29 +7,63 @@
...
+            choices=["series", "interview", "audio", "all"],
+            help="The type of items to bulk export, choices are: series, interview, assets and all",
         )
 
     def handle(self, *args, **options):
 
-        item_id = options["item_id"]
-        create_mods_records(item_id)
+        if options["item_id"]:
+            for item_id in options["item_id"]:
```
akohler, line 34: See previous comment about supporting multiple single item exports.  If this is desired, please refactor so the driver for this can reuse the `try/except` and the `count` output message currently in the bulk export.

oh_staff_ui/management/commands/create_mods_records.py
```diff
@@ -7,29 +7,63 @@
...
+
+        if options["bulk"]:
+            self._create_bulk_mods_records(options["bulk"])
+
+    def _create_mods_record(self, item_id: int) -> None:
+        """Given an item_id write the related MODS record to public location"""
+        try:
+            pi = ProjectItem.objects.get(id=item_id)
+            OralHistoryMods(pi).write_mods_record()
+            logger.info(f"Item: {pi.id} with ark {pi.ark} MODS record written")
```
akohler, line 45: Logging is good... but when doing the whole set, it quickly fills up the log, which currently defaults to showing the 200 most recent lines.  If it's useful to know it succeeded, switch this to `logger.debug` so it will only be in our dev environments (once I finish my PR...).  However, I think logging the rare errors is enough, instead of the common successes - that's what I'm doing with my other logging cleanup.

akohler (changes requested):
@kjallen This works well, other than the use of both bulk and individual switches which I think we should not support.  I also recommend allowing just a single item, rather than any number of them, via the `-i` switch - but if that behavior is useful, please refactor to provide the same sort of output for both modes.

oh_staff_ui/management/commands/create_mods_records.py
kjallen: Noted, this was more of a nice to have but not required. Removing.

oh_staff_ui/management/commands/create_mods_records.py
kjallen: I am switching to `logger.debug` and will keep in, I think useful for debugging in local environment.

akohler (approved):

PR #100, opened 2023-06-15
Sys 1308/oai envelope

kjallen:
This implements:
- SYS-1308 : Add OAI envelope
- SYS-1310 : Add GetRecord url option
- SYS-1311 : Add ListRecords url option

This replaces the core functionality of providing OH content via an OAI like endpoint, in order to deprecate the fragile current dldataprovider app and prevent the need for use of a stand alone OAI provider. 

Some feedback is appreciated, as some of the django best practices have not been followed, rather the OAI format for urls, but since there is only 1 client, I think we have some flexibility here.

A url parameter `verb` is required in the query string, this is the only required parameter, though it does not make sense without an `identifier` parameter as well for `GetRecord`

Example of use in the previous system:
https://webservices.library.ucla.edu/dldataprovider/oai2_0.do?verb=GetRecord&identifier=21198-zz002kppdz&metadataPrefix=mods

Matching equivalent in the new application:
http://localhost:8000/oai/?verb=GetRecord&identifier=21198-zz002kppdz

For all completed records, previous system:
https://webservices.library.ucla.edu/dldataprovider/oai2_0.do?verb=ListRecords&set=oralhistory&metadataPrefix=mods

Matching equivalent in new application:
http://localhost:8000/oai/?verb=ListRecords

Since the only metadata standard we are supporting is mods, and the only set we have is oral history, these parameters have been dropped from being required.

An ark that is empty or returns no result is returned with empty xml, consistent with the previous application.

Because only other machines should be reading this, any errors or exceptions encountered result in a `HttpResponseBadRequest()` rather than a user readable message.

Please review and provide feedback for improvement suggestions, particularly around bad input or error handling.

 ---
Feedback notes:

Valid behavior should be as follows when using:

A `verb` parameter must be supplied, and must be either `GetRecord` or `ListRecords`, any other values for verb, or absence of a verb  should return HttpResponseBadRequest()

If `verb=GetRecord` an `identifier` parameter must be supplied. If an `identifier` parameter is not supplied, an xml response will be returned but with `<error code="badArgument"/>` inside the `OAI` tag

If a `identifier` is supplied, if an `ark` is found for that value, a MODS record is returned wrapped in OAI elements. If no `ProjectItem` is found matching this `ark`, an empty `GetRecord` element is returned.

A test for these 3 cases (`GetRecord` with and without `identifier`, `ListRecords`) has been added.

Additional changes:

- `ark_ns` property added to `ProjectItem` for returning `ark`s with a dash replacing slash 
- Removed unused import in `OralHistoryMods`
- Changed behavior of `GetRecord` in case of no `identifier`
- Improved variable names in `OralHistoryMods`
- Added comments to `urls.py` to explain query parameter expectations of `/oai`
- Moved oai functions out of `OralHistoryMods` and into `views_util`
- Added more specific exception in `views` for `oai` path

---

Feedback notes continued

The feedback below has been incorporated and some more 'not ideal' changes to tests to get through a blocker.

- Addressing the http error code vs OAI response inconsistency, the oai view now does not return http error codes previous errors but follows the best practices requested for the errors encountered during an OAI request:
http://www.openarchives.org/OAI/openarchivesprotocol.html#ErrorConditions

To confirm error conditions:
- BadVerb (not GetRecord or ListRecords): http://localhost:8000/oai/?verb=GetRecordz&21198/zz000905jr
- BadArgument (Missing required parameter) : http://localhost:8000/oai/?verb=GetRecord

- The duplicate and confusing methods for `get_record` and `get_listrecords` have been squashed into 1 `get_records_oai` method per suggestion
- Inconsistencies with method parameters updated
- Unused imports removed

Tests:
- Some tests reworked to account for new error values
- References to `MODS.is_valid()` removed for now until a reliable method of accessing a remote XSD (or alternative) is implemented

### DIFF
--- a/oh_staff_ui/tests.py
+++ b/oh_staff_ui/tests.py
@@ -52,6 +52,7 @@ from oh_staff_ui.classes.OralHistoryMods import OralHistoryMods
 from oh_staff_ui.views_utils import (
     get_records_oai,
     get_bad_arg_error_xml,
+    get_bad_verb_error_xml,
 )
 
 
@@ -1160,33 +1161,12 @@ class ModsTestCase(TestCase):
 
         return ohmods
 
-    def test_valid_series_item_mods(self):
-        item = self.series_item
-        ohmods = OralHistoryMods(item)
-        ohmods.populate_fields()
-
-        self.assertEqual(ohmods.is_valid(), True)
-
-    def test_valid_interview_item_mods(self):
-        ohmods = self.get_mods_from_interview_item()
-
-        self.assertEqual(ohmods.is_valid(), True)
-
-    def test_valid_audio_item_mods(self):
-        item = self.audio_item
-        ohmods = OralHistoryMods(item)
-        ohmods.populate_fields()
-
-        self.assertEqual(ohmods.is_valid(), True)
-
     def test_valid_abstract_parse(self):
         ohmods = self.get_mods_from_interview_item()
 
         ohmods_from_string = load_xmlobject_from_string(
             ohmods.serializeDocument(), mods.MODSv34
         )
-
-        self.assertEqual(ohmods_from_string.is_valid(), True)
 
         # Assert abstract generated properly
         mods_xml = ohmods_from_string.serialize(pretty=True)
@@ -1198,9 +1178,6 @@ class ModsTestCase(TestCase):
         ohmods_from_string = load_xmlobject_from_string(
             ohmods.serializeDocument(), mods.MODSv34
         )
-
-        # Assert MODS still valid as read from string
-        self.assertEqual(ohmods_from_string.is_valid(), True)
 
         # Assert specific description type properly exists
         mods_xml = ohmods_from_string.serialize(pretty=True)
@@ -1218,21 +1195,18 @@ class ModsTestCase(TestCase):
 
     def test_valid_mods_created_date(self):
         ohmods = self.get_mods_from_interview_item()
-        self.assertEqual(ohmods.is_valid(), True)
         self.assertTrue(
             b"<mods:dateCreated>2000</mods:dateCreated>" in ohmods.serializeDocument()
         )
 
     def test_valid_mods_alttitle(self):
         ohmods = self.get_mods_from_interview_item()
-        self.assertEqual(ohmods.is_valid(), True)
         self.assertTrue(
             b'<mods:titleInfo type="alternative">' in ohmods.serializeDocument()
         )
 
     def test_valid_mods_altid(self):
         ohmods = self.get_mods_from_interview_item()
-        self.assertEqual(ohmods.is_valid(), True)
         self.assertTrue(
             b'<mods:identifier type="OPAC">Alt Id</mods:identifier>'
             in ohmods.serializeDocument()
@@ -1330,9 +1304,13 @@ class ModsTestCase(TestCase):
         bad_response = get_bad_arg_error_xml("GetRecordWithoutIdentifier")
         self.assertTrue(b'<error code="badArgument"/>' in bad_response)
 
+    def test_bad_getrecord_request(self):
+        bad_response = get_bad_verb_error_xml("BadVerb")
+        self.assertTrue(b'<error code="badVerb"/>' in bad_response)
+
     def test_getrecord_request(self):
         id_to_check = self.series_item.ark
-        response = get_records_oai(id_to_check)
+        response = get_records_oai(verb="GetRecord", ark=id_to_check)
         id_tag = f'identifier="{id_to_check}"'
 
         self.assertTrue(bytes(id_tag, "utf-8") in response)
--- a/oh_staff_ui/tests.py
+++ b/oh_staff_ui/tests.py
@@ -1184,7 +1184,7 @@ class ModsTestCase(TestCase):
         ohmods = self.get_mods_from_interview_item()
 
         ohmods_from_string = load_xmlobject_from_string(
-            ohmods.serializeDocument(), mods.MODS
+            ohmods.serializeDocument(), mods.MODSv34
         )
 
         self.assertEqual(ohmods_from_string.is_valid(), True)
@@ -1197,7 +1197,7 @@ class ModsTestCase(TestCase):
         ohmods = self.get_mods_from_interview_item()
 
         ohmods_from_string = load_xmlobject_from_string(
-            ohmods.serializeDocument(), mods.MODS
+            ohmods.serializeDocument(), mods.MODSv34
         )
 
         # Assert MODS still valid as read from string
--- a/oh_staff_ui/tests.py
+++ b/oh_staff_ui/tests.py
@@ -49,6 +49,11 @@ from oh_staff_ui.models import (
 from oh_staff_ui.classes.OralHistoryFile import OralHistoryFile
 from oh_staff_ui.classes.AudioFileHandler import AudioFileHandler
 from oh_staff_ui.classes.OralHistoryMods import OralHistoryMods
+from oh_staff_ui.views_utils import (
+    get_listrecords_oai,
+    get_record_oai,
+    get_bad_arg_error_xml,
+)
 
 
 class MediaFileTestCase(TestCase):
@@ -1321,6 +1326,23 @@ class ModsTestCase(TestCase):
         # Verify file exists
         p = Path(f"{settings.MEDIA_ROOT}/{settings.OH_STATIC}/mods/{ark_ns}-mods.xml")
         self.assertTrue(p.is_file())
+
+    def test_bad_getrecord_request(self):
+        bad_response = get_bad_arg_error_xml("GetRecordWithoutIdentifier")
+        self.assertTrue(b'<error code="badArgument"/>' in bad_response)
+
+    def test_getrecord_request(self):
+        id_to_check = self.series_item.ark
+        response = get_record_oai(id_to_check)
+        id_tag = f'identifier="{id_to_check}"'
+
+        self.assertTrue(bytes(id_tag, "utf-8") in response)
+
+    def test_listrecords_request(self):
+        response = get_listrecords_oai("ListRecords")
+        self.assertTrue(
+            b'<request metadataPrefix="mods" verb="ListRecords">' in response
+        )
 
 
 class FileMetadataMigrationTestCase(SimpleTestCase):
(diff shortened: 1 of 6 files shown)


oh_staff_ui/views_utils.py
```diff
@@ -466,3 +470,50 @@ def run_process_file_command(
...
+    pi_set = ProjectItem.objects.filter(status__status__iexact="completed")
+
+    if ark:
+        ark = ark.replace("-", "/")
+        pi_set = pi_set.filter(ark=ark)
+
+    verb_element = etree.Element(verb)
+
+    for pi in pi_set:
+        verb_element.append(OralHistoryMods(pi).add_oai_envelope())
```
akohler, line 486: Does the `OralHistoryMods` class need to provide records both with and without an OAI envelope?  If not, or if the norm is with envelope, I think the class should just provide that output automatically, without this external routine telling the class to add the envelope.

oh_staff_ui/views_utils.py
```diff
@@ -466,3 +470,50 @@ def run_process_file_command(
...
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+                xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/ 
+                http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd" />
+            """
+    oai_tree = etree.fromstring(oai_envelope)
+    d = datetime.now()
+
+    etree.SubElement(oai_tree, "responseDate").text = d.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    req = f"""<request metadataPrefix="mods"
```
akohler, line 508: Please add comments explaining this.  I think the `reverse` here is a Django URL shortcut, so... is some sort of URL being built here?

oh_staff_ui/urls.py
```diff
@@ -17,4 +17,5 @@
     path("upload_file/<int:item_id>", views.upload_file, name="upload_file"),
     path("order_files/<int:item_id>", views.order_files, name="order_files"),
     path("browse/", views.browse, name="browse"),
+    path("oai/", views.oai, name="oai")
```
akohler, line 20: Since the logic for this is all in the view, please add comment(s) here to document that `/oai/` needs to be called with `verb` and optionally `ark` parameters.

oh_staff_ui/views.py
```diff
@@ -192,3 +195,24 @@ def order_files(request: HttpRequest, item_id: int) -> HttpResponse:
...
+
+        if verb == "GetRecord":
+            xml_content = get_record_oai(ark)
+
+        elif verb == "ListRecords":
+            xml_content = get_listrecords_oai("ListRecords")
+
+        return HttpResponse(xml_content, content_type="text/xml")
+
+    except Exception as e:
```
akohler, line 217: The `as e` is not needed, since never used.  See also comment above about changing to a more specific error.

oh_staff_ui/views.py
```diff
@@ -192,3 +195,24 @@ def order_files(request: HttpRequest, item_id: int) -> HttpResponse:
...
+
+
+def oai(request: HttpRequest) -> HttpResponse:
+    try:
+        # Verb is required, ark is optional
+        verb = request.GET["verb"]
+        ark = request.GET.get("identifier")
+
+        if verb not in ("GetRecord", "ListRecords"):
+            raise Exception
```
akohler, line 207: Change this from generic Exception, maybe `ValueError` or `HTTPError`

akohler (changes requested):
@kjallen I've added several comments requesting specific changes.  A few may reflect my lack of knowledge about OAI; happy to meet to talk through these if they don't make sense.

In addition, please add test(s) - there should be at least one for each verb, using the sample records you're already creating for other tests.  Since those tests already confirm the MODS output is correct, I think just tests to confirm the OAI wrappers may be enough.

> A url parameter verb is required in the query string, this is the only required parameter, though it does not make sense without an identifier parameter as well for GetRecord

Should there be code added to handle this case?  Either require an identifier parameter for `GetRecord`, or return an empty response, or ?

`curl -s "http://localhost:8000/oai/?verb=GetRecord"` took about a minute to run and returned an XML document with 4700 records.

The output from that is almost identical to that from `curl -s "http://localhost:8000/oai/?verb=ListRecords"`.  After formatting, the differences:
```
$ diff GetRecord.xml ListRecords.xml
3,4c3,4
<   <responseDate>2023-06-15T17:25:42Z</responseDate>
<   <request metadataPrefix="mods" verb="GetRecord">/oai/
---
>   <responseDate>2023-06-15T17:36:17Z</responseDate>
>   <request metadataPrefix="mods" verb="ListRecords">/oai/
6c6
<   <GetRecord>
---
>   <ListRecords>
161586c161586
<   </GetRecord>
---
>   </ListRecords>
```
So it seems like `GetRecord` with no identifier should be handled differently - unless this is what the standard requires.

Thanks --Andy

akohler:
@kjallen wrote
> Some feedback is appreciated, as some of the django best practices have not been followed, rather the OAI format for urls, but since there is only 1 client, I think we have some flexibility here.

Django's fine with "normal" URL GET parameters like OAI uses, and pulling them out of the `request.GET` dictionary as your view does is appropriate.

oh_staff_ui/views_utils.py
```diff
@@ -8,6 +10,7 @@
 from django.db.models import CharField, Model, Q, QuerySet
 from django.forms import BaseFormSet, Form, formset_factory
 from django.http.request import HttpRequest  # for code completion
+from django.urls import reverse
```
akohler, line 13: This import is no longer used.

oh_staff_ui/views_utils.py
```diff
@@ -466,3 +470,105 @@ def run_process_file_command(
...
+    e_req = etree.fromstring(req)
+    e_req.set("verb", verb)
+
+    if ark:
+        e_req.set("identifier", ark)
+
+    return e_req
+
+
+def get_bad_arg_error_xml(verb: str, req_url: str = None) -> str:
```
akohler, line 541: Please document this method.  I see it adds a bit of XML, but there's no context here; I have to go back to the `oai` view to see it gets called when `GetRecord` verb is used with no ARK.  More information would be helpful, like why do this (returning HTTP 200 with this error information in the XML) rather than returning an HTTP 400 or similar.

oh_staff_ui/views_utils.py
```diff
@@ -466,3 +470,105 @@ def run_process_file_command(
...
+    oai_envelope = get_oai_envelope()
+
+    oai_envelope.append(get_response_date_element())
+    oai_envelope.append(get_request_element(verb, req_url=req_url))
+
+    error_str = f"""<error code="badArgument"/>"""
```
akohler, line 547: `flake8` points out this f-string is not needed since there's no placeholder (or, the expected placeholder is missing).

oh_staff_ui/views_utils.py
```diff
@@ -466,3 +470,105 @@ def run_process_file_command(
     # testing confirms this, though it does close after 30 seconds or so.
     # To be safe, explicitly close this one when done.
     connection.close()
+
+
+def get_record_oai(ark: str = None, req_url: str = None) -> str:
```
akohler, line 475: Please add comments here & in `get_listrecords_oai()`.  Developers who don't know OAI - like me - need more context.  I think that the `GetRecord` verb is just a specialized case of `ListRecords`, right - retrieving a "list" of one specific record?  If so, based on the code in the `oai` view, it seems like just one of these 2 methods is needed.  Maybe just `get_records_oai(verb, ark, req_url)`?

But if 2 separate methods is preferable / necessary, add comments to explain the relation and the differences.

oh_staff_ui/views.py
```diff
@@ -192,3 +196,29 @@ def order_files(request: HttpRequest, item_id: int) -> HttpResponse:
...
+        ark = request.GET.get("identifier")
+
+        if verb not in ("GetRecord", "ListRecords"):
+            raise ValueError
+
+        req_url = request.build_absolute_uri("?")
+
+        if verb == "GetRecord":
+            if ark:
+                xml_content = get_record_oai(ark, req_url=req_url)
```
akohler, line 214: Is `req_url=req_url` needed here, as opposed to just `req_url` like in line 216?  I think these can and should be consistent.

akohler (changes requested):
@kjallen Thanks for making changes I previously requested.  There are a few more minor changes to make, mostly documentation improvements to help understand what some of the OAI code does, and why.

More troubling: at present, 8 of the tests fail.  All are MODS validation-related, one of these:
```
    self.assertEqual(ohmods.is_valid(), True)
    self.assertEqual(ohmods_from_string.is_valid(), True)
```
I asked Zoe to pull and run the tests too, and she saw the same 8 errors.

I don't know why these are happening, or why they didn't happen for you.  I can't find recent code changes which seem relevant.  I even deleted my Docker image, checked out old code which I know had passing tests - commit `e025aa354136093a936a16b94133f91516ffc43d` from May 30 - and rebuilt my image from that code... the same tests still fail.  Something upstream maybe?

oh_staff_ui/tests.py
```diff
@@ -1322,6 +1300,27 @@ def test_writing_single_mods(self):
         p = Path(f"{settings.MEDIA_ROOT}/{settings.OH_STATIC}/mods/{ark_ns}-mods.xml")
         self.assertTrue(p.is_file())
 
+    def test_bad_getrecord_request(self):
+        bad_response = get_bad_arg_error_xml("GetRecordWithoutIdentifier")
+        self.assertTrue(b'<error code="badArgument"/>' in bad_response)
+
+    def test_bad_getrecord_request(self):
```
akohler, line 1307: `test_bad_getrecord_request()` is defined twice, with different code.

oh_staff_ui/views.py
```diff
@@ -7,6 +7,7 @@
 from django.contrib.auth.decorators import login_required
 from django.http.request import HttpRequest  # for code completion
 from django.http.response import HttpResponse  # for code completion
+from django.http.response import HttpResponseBadRequest
```
akohler, line 10: Remove this unused import.

akohler (changes requested):
@kjallen 2 more minor changes - an unused import, and a duplicate test method name - then this can be merged.  
All tests now pass for me.  The added comments are helpful for understanding the OAI-specific implementations.
Thanks --Andy

akohler (approved):

PR #102, opened 2023-08-08
SYS-1369: Increase form field limit for description text

akohler:
Implements [SYS-1369](https://jira.library.ucla.edu/browse/SYS-1369), while bringing two other minor issues along:
* [SYS-1345: Prevent fixtures from being reloaded in production](https://jira.library.ucla.edu/browse/SYS-1345)
* [SYS-1234: Display application version (via release notes)](https://jira.library.ucla.edu/browse/SYS-1234)

This PR fixes a problem where the item metadata form limited Description text entry to 1024 characters, instead of allowing the full 8000 characters the model supports.  Migrated Description data includes values nearly 4000 characters long, and Oral History staff occasionally need to add very long tables of contents.

Since this was a very minor technical change which needed to be done, I combined it with two other minor issues which might otherwise not get done.
1) Running the local Docker environment while connecting to the remote production database would cause fixtures to be (re)loaded, which could restore the corresponding production data to its default state.  This was fixed by changing `docker-compose_REMOTE.yml` to set `DJANGO_RUN_ENV=prod`, explicitly using the production environment.
2) There was no easy way to see the deployed application version, and no central location for release notes telling users what had changed.  I combined these by adding `/release_notes/`, which include both version and notes.  I also added a note to `README.md` to document the release process.  There's also a link to "Release Notes" in the application's top menu bar.

#### Testing
There's not much to test.  Start the application, create or edit a record and confirm that long text can be added to the Description field.  http://127.0.0.1:8000/item/10334 is an existing example which has about 3800 characters; copy and paste it into the field again to verify 7500+ characters can be entered, stored and displayed.

Visit http://127.0.0.1:8000/release_notes/ to confirm that release notes display, including version information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,3 +278,12 @@ $ docker-exec exec django python manage.py shell
 (1, {'oh_staff_ui.MediaFile': 1})
 ```
 Deleting a `MediaFile` object does _*not*_ currently delete any `FileField` object associated with it, so it's necessary to do a `file.delete()` first.
+
+#### Preparing a release
+
+Our deployment system is triggered by changes to the Helm chart.  Typically, this is done by incrementing `image:tag` (on or near line 9) in `charts/prod-ohstaff-values.yaml`.  We use a simple [semantic versioning](https://semver.org/) system:
+* Bug fixes: update patch level (e.g., `v1.0.1` to `v1.0.2`)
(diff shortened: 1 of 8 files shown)


ztucker4 (approved):
Hi @akohler, this looks good. I've confirmed that I can create/edit Descriptions up to 8000 characters now, the release notes display correctly, and the docker-compose production changes make sense.
Merging now, thanks!

PR #105, opened 2023-10-12
Added OAI documentation

kjallen:
Addition of OAI documentation, request to read and confirm it makes sense, catch any typos or formatting issues, and requests for additional clarification.


akohler (approved):
@kjallen Seems clear enough to me, and matches the implementation in code.  There's a risk of code / documentation diverging with this documented outside of the code, so if we have to modify the `OralHistoryMods` implementation in the future I might remove details like the subjects, but that's a future problem... it's fine for now, merging, thanks.

PR #108, opened 2024-03-14
Timed log attribute added for TEI XML transcripts with test

kjallen:
The public site expects TEI transcripts to have a `mods:location/mods:url[@usage="timed log"` attribute present in order to trigger specific functionality. This was removed in the initial implementation -- this commit adds it back.

This attribute value is not part of the [MODS specification](https://www.loc.gov/standards/mods/userguide/attributes.html#usage) for the location element, so no validation against the schema is done, either in tests or the application.

A minor change was made to the `OralHistoryMods` class, along with an additional test which confirms if a TEI transcript (filetype of `text_master_transcript`) is present, a `<mods:url usage="timed log">` tag is also present.

To test:
- Run tests, all should pass
- Generating an OAI record of an item that contains a TEI xml media file will output a `usage="timed log"` attribute/value. An example record can be found [here](http://localhost:8000/oai/?verb=GetRecord&identifier=21198/zz0008zmvt)

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v1.0.6
+  tag: v1.0.7
   pullPolicy: Always
 
 nameOverride: ""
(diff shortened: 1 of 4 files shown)


akohler (approved):
@kjallen These code changes and the test look fine to me.  I reviewed the sample MODS for `21198/zz0008zmvt` and the record in the staff UI, and confirmed the attribute is set appropriately given the data.

The only thing missing was an update to `release_notes.html` to correspond to the image tag change, so I added that.

Merging now, should be live in the next few minutes.

PR #111, opened 2024-04-03
Adjust allowed_hosts value to be a comma-separated string

sgurnick:
We are provided a list of host names to the `allowed_hosts` value and it appears the way we were passing these to the configmap template using the `range` loop was not working correctly. 

Since Django is expecting a comma-separated list in projects.yaml, we'll just provide that in the values file.


akohler (approved):
@sgurnick Looks good to me, since this is what Django ultimately needs, might as well pass it that way from the beginning.  Merging.

PR #113, opened 2024-04-05
SYS-1576: "View on public site" links

sourcefilter:


charts/templates/configmap.yaml
```diff
@@ -22,4 +22,4 @@ data:
   DJANGO_OH_MASTERSLZ: {{ .Values.django.env.oh_masterslz }}
   DJANGO_OH_STATIC: {{ .Values.django.env.oh_static }}
   DJANGO_OH_WOWZA: {{ .Values.django.env.oh_wowza }}
-  
\ No newline at end of file
+  DJANGO_OH_PUBLIC_SITE: {{ .Values.django.env.oh_public_site }}
```
akohler, line 25: Changing `configmap.yaml` to add a new value is appropriate,  However, there's one more change needed when doing this: the version of the Chart itself (the big-C `Chart.yaml`) also needs to be changed, when anything other than the specific values in `prod-<appname>-values.yaml` change.

Please increment `version` in `Chart.yaml`.  This typically gets bumped by a minor version, so in this case, from `0.4.0` to `0.5.0`.

There's a little more about this in our ["cookbook"](https://uclalibrary.atlassian.net/wiki/x/T075B), the final section of "Goal: Update environment variables only".

oh_staff_ui/views_utils.py
```diff
@@ -1,4 +1,6 @@
 import logging
+import urllib.parse
```
akohler, line 2: This import can be removed with the (proposed) code change at line 231, below.

oh_staff_ui/views_utils.py
```diff
@@ -214,9 +216,25 @@ def get_edit_item_context(item_id: int) -> dict:
...
 
+def get_public_site_url(item: ProjectItem) -> str:
+    if not (
+        settings.OH_PUBLIC_SITE
+        and item.status.status in ("Completed", "Completed with minimal metadata")
+    ):
+        return None
+
+    if item.type.type == "Series":
+        return f"{settings.OH_PUBLIC_SITE}/?f%5Bseries_facet%5D%5B%5D={urllib.parse.quote_plus(item.title)}"
```
akohler, line 231: Instead of building a pre-encoded/escaped URL here, I prefer the simpler and self-documenting:

`return f"{settings.OH_PUBLIC_SITE}/?f[series_facet][]={item.title}"`

This string is being passed to a Django template, which can escape/encode URLs when needed (though I tested, and this doesn't seem to need it).  Please try it yourself, and if you agree, let's use this simpler non-encoded version.

If you do want to keep the pre-encoded version, please document it via a comment above that line, so readers know what's been encoded.

akohler (changes requested):
@sourcefilter Very good, with just a few minor changes needed.  I tested with several items and the link appeared (or didn't) as expected, and worked correctly to retrieve interview records and to do series-level searches.

From my testing, it's not necessary to encode the series-level faceted searches into the public site.  That doesn't prove it's never required, so I won't argue if you prefer the encoding - but if we keep it, please document it in the method for future reference.

Speaking of tests, I think this warrants a small new test case being added to `tests.py`, with a few tests to confirm that public URLs are generated / not generated when appropriate, and that the formats of the generated URLs are correct.

Finally, for future pull requests, please add a description following our preferred format.  There are a couple of examples linked on our [best practices page](https://uclalibrary.atlassian.net/wiki/x/k8L4B), or take a look at #92 for an example from this project.

Thanks --Andy K.

akohler (approved):

akohler:
@sourcefilter The changes look good, especially the new tests - comprehensive, and great use of `subTest` which I haven't used before but definitely will in future.

You asked about testing whether the new link is present in the rendered template.  I think that's not necessary for this project - nothing wrong with including such a test in future, but the relevant logic in `get_public_site_url()` is what matters, and you've added tests which cover creation and content, including the valid `None` option.  The template just shows it if it's not `None`, which I view as core Django functionality which we don't need to test.

Merging now, thanks.

PR #115, opened 2024-04-23
SYS-1594 Adjustment to mount the OH masters volume

sgurnick:
The changes here allow for mounting both the original OH Master Landing Zone volume and the new OH Masters volume. 

Once we confirm using the new OH Masters volume is working as-expected, we can remove the Landing Zone volume. 

@akohler Please review these changes when you get a chance, in particular the changes I made in `prod-ohstaff-values.yaml` and the `configmap.yaml`.

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -56,6 +56,7 @@ django:
     # Where Django looks for uploaded files
     oh_media_root: "/media"
     # Relative paths within oh_media_root, used by Django
+    oh_masters: "oh_masters"
     oh_masterslz: "oh_lz"
     oh_static: "oh_static"
     oh_wowza: "oh_wowza"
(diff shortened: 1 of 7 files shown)


charts/prod-ohstaff-values.yaml
```diff
@@ -56,6 +56,7 @@ django:
     # Where Django looks for uploaded files
     oh_media_root: "/media"
     # Relative paths within oh_media_root, used by Django
+    oh_masters: "oh_masters"
```
akohler, line 59: I expect to see this added to the generic `values.yaml` as well, for consistency.  It may not strictly be necessary but it's what we do for other specific values.

charts/prod-ohstaff-values.yaml
```diff
@@ -78,6 +79,11 @@ persistentVolumes:
     server: "lib-partners.in.library.ucla.edu"
     path: "/OralHistory"
   ohmasters:
+    storage: "7Ti"
```
akohler, line 82: Similar to the above, I expect to see both `ohmasters` and `ohmasterslz` placeholders in the generic `values.yaml`.

akohler (changes requested):
@sgurnick I think this would work - the important things look correct, or at least consistent.  But for full consistency, I think `values.yaml` should be updated to have placeholders and default values matching the production values file.

Thanks --Andy

akohler (approved):

akohler:
@sgurnick With the default values added, looks good - merged.

PR #119, opened 2024-04-26
SYS-1522: Exclude series records from OAI feed

sourcefilter:


akohler (approved):
@sourcefilter The code and tests look good.  I renamed one test to conform to [PEP8 conventions](https://peps.python.org/pep-0008/#function-and-variable-names)" `test_series_records_NOT_included()` -> `test_series_records_are_excluded()`

In addition to the automated tests, I pulled before and after versions of the full OAI feed from my local system and manually confirmed that Series records which were included before are not included now, and that the change did not affect the expected occurrences of the series as a related item entry.

Merging.

PR #122, opened 2024-05-02
SYS-1601: add reprocess_derivative_images command

ztucker4:
Implements [SYS-1601](https://uclalibrary.atlassian.net/browse/SYS-1601)

Adds a new Django management command, `reprocess_derivative_images`. This command takes one optional argument, `--project_item_id`. If supplied, only images associated with the specified ProjectItem will be reprocessed. Otherwise, all images will be processed. Example usage: ```docker-compose exec django python manage.py reprocess_derivative_images --project_item_id 10331```

This command simply creates new derivative images from all master images instead of attempting to determine which ones require correction. This is simpler, and has the advantage that the new derivative images are now correctly associated with their respective masters through the `parent` field. The "created by" metadata is altered for the derivatives, but remains intact for masters.

My testing was limited to files I uploaded myself, since the original files must actually exist in the expected media directory. Please check if the path to access original master files (line 29 of `reprocess_derivative_images.py`) will work in production - I'm a bit unsure of this.

Finally, this PR also includes:
* A tiny change to `views.py` to sort the file information on the file upload page by sequence number. The default sorting (chronological?) was inconvenient after reprocessing a ProjectItem with multiple masters.
* Release notes and tag bump for deployment.

[SYS-1601]: https://uclalibrary.atlassian.net/browse/SYS-1601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oh_staff_ui/management/commands/reprocess_derivative_images.py
+++ b/oh_staff_ui/management/commands/reprocess_derivative_images.py
@@ -20,8 +20,9 @@ def reprocess_derivative_images(master_images: QuerySet, request: HttpRequest) -
     """Reprocess all derivative images from masters."""
     try:
         for master_image in master_images:
-            logger.info(f"Processing master image {master_image.file.name}.")
-            # Get id of newly created MediaFile to use as parent for derivative(s)
+            logger.info(
+                f"Recreating derivatives for master image {master_image.file.name}."
+            )
             master_id = master_image.id
             # Create an OHF for the master file so we can use ImageFileHandler
             master_file = OralHistoryFile(
@@ -63,6 +64,16 @@ def reprocess_derivative_images(master_images: QuerySet, request: HttpRequest) -
     except (CommandError, ValueError) as ex:
         logger.error(ex)
         raise
+    finally:
+        # Delete the temporary files;
+        # don't throw FileNotFoundError if for some reason it doesn't exist.
+        try:
+            Path(submaster_file_name).unlink(missing_ok=True)
+            Path(thumbnail_file_name).unlink(missing_ok=True)
+        except UnboundLocalError:
+            # Swallow this, which happens when derivative creation fails
+            # so file_name variable is not defined.
+            pass
 
 
 def delete_existing_derivative_images(master_images: QuerySet) -> None:
@@ -92,15 +103,17 @@ def delete_existing_derivative_images(master_images: QuerySet) -> None:
                         mf.delete()
                         logger.info(f"Deleted MediaFile {mediafile_id} from database.")
                     else:
-                        logger.error(
-                            f"Failed to delete {file_name} from file system. Exiting."
+                        raise CommandError(
+                            f"Failed to delete {file_name} from file system."
                         )
-                        return
+
                 else:
-                    logger.error(
-                        f"File {file_name} does not exist in the file system. Exiting."
+                    logger.info(
+                        f"File {file_name}, for MediaFile {mediafile_id}, "
+                        "does not exist in the file system."
                     )
-                    return
+                    mf.delete()
+                    logger.info(f"Deleted MediaFile {mediafile_id} from database.")
 
             logger.info(
                 f"Finished deleting existing derivative images for project item {project_item}."
@@ -124,21 +137,20 @@ class Command(BaseCommand):
     def add_arguments(self, parser: CommandParser) -> None:
         parser.add_argument(
             "--project_item_id",
-            type=int,
-            help="ID of ProjectItem to reprocess derivative images for",
-            required=False,
+            type=str,
+            help="ID of ProjectItem to reprocess derivative images for, or 'ALL' to reprocess all.",
+            required=True,
         )
 
     def handle(self, *args, **options) -> None:
-        if options["project_item_id"]:
-            master_images = MediaFile.objects.filter(
-                item__id=options["project_item_id"],
-                file_type__file_code="image_master",
-            ).order_by("id")
-
-        else:
+        if options["project_item_id"] == "ALL":
             master_images = MediaFile.objects.filter(
                 file_type__file_code="image_master"
+            ).order_by("id")
+        else:
+            master_images = MediaFile.objects.filter(
+                item__id=int(options["project_item_id"]),
+                file_type__file_code="image_master",
             ).order_by("id")
 
         master_image_count = master_images.count()
--- a/oh_staff_ui/management/commands/reprocess_derivative_images.py
+++ b/oh_staff_ui/management/commands/reprocess_derivative_images.py
@@ -1,0 +1,115 @@
+import logging
+from pathlib import Path
+from django.core.management.base import BaseCommand, CommandParser
+from django.core.management.base import CommandError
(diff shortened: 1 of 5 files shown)


oh_staff_ui/views.py
```diff
@@ -132,7 +132,7 @@ def show_log(request, line_count: int = 200) -> HttpResponse:
 @login_required
 def upload_file(request: HttpRequest, item_id: int) -> HttpResponse:
     item = ProjectItem.objects.get(pk=item_id)
-    files = MediaFile.objects.filter(item=item)
+    files = MediaFile.objects.filter(item=item).order_by("sequence")
```
akohler, line 135: Adding a sort order is good, thanks - I don't know why we didn't do this originally.  Now that I see this, one improvement needed: files with the same sequence seem to sub-sort randomly, or maybe by an internal id.  See for example http://127.0.0.1:8000/upload_file/2164 (once production data is loaded locally), where sometimes a master is before submaster, and sometimes after.  

There's no perfect solution - I think we can't trust `MediaFile.create_date()` completely, for example.  But happily, `master` -> `submaster` -> `thumbnail`... I did some testing, and `order_by("sequence", "file_type__file_code")` seems to work well... when the data is correct.

Some migrated data, like for item 1420, is bad.  We had conflicts between file names and locations and DLCS file usage / types.  But for new data this sub-ordering should work, and at some point we may do similar data cleanup projects for non-images.

akohler (changes requested):
@ztucker4 Some requested changes, especially around file deletion.  I have not yet tried actually running the command; I'll wait to do that until the deletion is improved.

I'm really paranoid about deletion, so I'd like to see more logging of data about the files being deleted, and the new files being created.  One concern is that, using locally-created files - which due to your thumbnail fix, are already correct - there's no proof being logged that files are actually deleted and that the new files are different (in some cases, anyhow).  So please think about (a) what's meaningful to log, and (b) how we can prove that the old files are really being deleted and replaced.

It's awkward to test management commands, and I know much of the code in this one is copied from classes which do have at least some tests.  But I think we should have tests for the `delete_existing_derivative_images()`  and `reprocess_derivative_images()` methods in this command.  See [this article](https://adamj.eu/tech/2020/09/07/how-to-unit-test-a-django-management-command/) for some simple examples of testing methods from a management command - I think we don't need to test the command itself (via `call_command` and parsing the output, just the relevant methods within it.

Thanks --Andy

ztucker4:
Hi @akohler, this is finally ready for another look. Fixes include:
* Updated `delete_existing_derivative_images()` to match on `ProjectItem` instead of `MediaFile.parent`. This is possibly a broader stroke than needed, but it's simple and the command is intended to be used on all master images anyway.
* Derivative image files on disk are now deleted alongside their `MediaFile`s in the database.
* More logging is added throughout `reprocess_derivative_images.py`, especially within the deletion method. There are now line-by-line updates about both file and database deletion operations, and errors that halt the command if a deletion fails.
* Testing! Only two new test cases, which cover deletion and reprocessing of freshly-processed master images. Please let me know if I should add more cases here to cover other cases.
* Added additional `file_code` sort to the `upload_file` view.

akohler (changes requested):
@ztucker4 This is getting closer.  I like the increased logging and the tests, and the "happy path" works as expected.

I think the deletion logic needs reworking, especially the cases where an error occurs - I don't think returning at those points is correct.  Details in specific comments.  

There are some smaller improvements as well, like preventing the script from running against the full set if no arguments are provided and cleaning up fully after tests and derivative regeneration.

To help with the manual part of testing, I wrote a shell script which resets a specific item to a consistent state, along with some corruption to simulate problems.  There's also a simple Django-specific Python script to do full file deletion (master and derivatives) for convenience.  Feel free to take a look: https://gist.github.com/akohler/cfa49def042c90dcca0ce0683b6b26f1

ztucker4:
@akohler, this is ready for yet another review. Issues addressed since yesterday:
* `--project_item_id` is now required, accepting either a single ID or "ALL"
* Updated logic in `delete_existing_derivative_images()`:
  * If the expected file does not exist for a `MediaFile`, log this but delete the `MediaFile` anyway (lines 110-116).
  * If the file exists, we attempt to delete it, and the file still exists, raise a `CommandError` (lines 105-108). I don't think this case should happen, as we should get an exception from the file deletion operation anyway like you mentioned. I haven't been able to simulate this case locally.
* Added `finally` block to `reprocess_derivative_images()` to delete temporary image files from `/tmp/`
* Tag value and release notes corrected
* Fixed capitalization and fixture usage in `test.py`
* Deleted incorrect comment in command file

oh_staff_ui/management/commands/reprocess_derivative_images.py
```diff
@@ -0,0 +1,160 @@
...
+                else:
+                    logger.info(
+                        f"File {file_name}, for MediaFile {mediafile_id}, "
+                        "does not exist in the file system."
+                    )
+                    mf.delete()
+                    logger.info(f"Deleted MediaFile {mediafile_id} from database.")
+
+            logger.info(
+                f"Finished deleting existing derivative images for project item {project_item}."
```
akohler, line 119: Please add the project item id to this message, for easier lookup / troubleshooting if needed - at present it just has the item title.  Something like "...for project item {project_item} ({project_item.id})" should be fine.

oh_staff_ui/management/commands/reprocess_derivative_images.py
```diff
@@ -0,0 +1,160 @@
...
+            pass
+
+
+def delete_existing_derivative_images(master_images: QuerySet) -> None:
+    """Delete existing derivative images from masters."""
+    try:
+        for master_image in master_images:
+            project_item = master_image.item
+            logger.info(
+                f"Deleting existing derivative images for project item {project_item}."
```
akohler, line 85: Please add the project item id to this message, for easier lookup / troubleshooting if needed - at present it just has the item title.  Something like "...for project item {project_item} ({project_item.id})" should be fine.

akohler (changes requested):
@ztucker4 Almost there!  The new deletion logic looks correct, as does making the item id argument mandatory.  Just a couple of logging tweaks to add the item id for extra info.

The main change needed is in the test's `tearDown()`, which is now too thorough.  Once that's fixed, I think this will be ready to go.

ztucker4:
Fixes made - added `ProjectItem` IDs in deletion logs and rewrote `tearDown()` based on your `delete_media_files()` example. Ready for (hopefully final) review!

akohler (approved):
@ztucker4 This all looks good now!  In addition to the single-item scenarios I've been testing locally, I also cleared out all local `MediaFile` data and added a variety of files to several items.  Running the command with `--project_item_id ALL` did what it should, rebuilding the image derivatives and not touching anything else.

The revised `tearDown()` for the tests also cleaned up what it should and nothing more.

Merging now, thanks.

PR #124, opened 2024-05-08
SYS-934: Fix master file access in production

akohler:
Fixes a problem with [SYS-934](https://uclalibrary.atlassian.net/browse/SYS-934).  Bumps version to `v1.1.1` for deployment.

This PR should fix a problem caused by my misunderstanding of how media (user-uploaded) files are handled by Django in production environments, when `DEBUG=False`.  It uses Django's built-in view `django.views.static.serve()` to serve files in `MEDIA_ROOT`... which is similar to what I was doing before, via a different built-in view which I thought was a short-cut for this functionality.  Apparently serving via `MEDIA_URL` is not allowed in production, but serving files from the place `MEDIA_URL` points to... is.

Along with this minor change in `urls.py`, I added a test which forces simulation of production mode `(RUN_ENV="prod", DEBUG=False)` and uses Django's built-in test client to confirm that a request for a media file succeeds (response status code 200).

I also started my local environment with those two environment variables set as above:
```
    environment:
      # Force the environment to be production, when connecting from Docker in this way.
      - DJANGO_RUN_ENV=prod
      - DJANGO_DEBUG=False
```
Master files were accessible in this environment.

Finally, I started my local environment in the normal way, and master files (and all other local files) were available as expected.

So, I'm hopeful....

[SYS-934]: https://uclalibrary.atlassian.net/browse/SYS-934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v1.1.0
+  tag: v1.1.1
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/templates/oh_staff_ui/release_notes.html
+++ b/oh_staff_ui/templates/oh_staff_ui/release_notes.html
@@ -4,6 +4,12 @@
 <h3>Release Notes</h3>
 <hr/>
 
+<h4>1.1.1</h4>
+<p><i>May 8, 2024</i></p>
+<ul>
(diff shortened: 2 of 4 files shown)


ztucker4 (approved):
Looks good to me. Tested as described in PR description (set environment variables to simulate both dev and prod), and accessing various types of master files works in both cases. New test also passes. Merging - fingers crossed this works!

PR #125, opened 2024-05-11
SYS-1598: allow deletion of derivative media files

ztucker4:
Implements [SYS-1598](https://uclalibrary.atlassian.net/browse/SYS-1598), which absorbed [SYS-1599](https://uclalibrary.atlassian.net/browse/SYS-1599)

This PR allows users to delete files (both Master and derivative) from item records. Users in the Oral History Staff user group will see a new column in the "File Information" table on `/upload_file/` pages containing a delete button for every file. When clicked, this button brings up a popup asking the user to confirm deletion of the selected file and its children (if any). If "Yes" is selected in this popup, the new `delete_file()` view is called and the relevant MediaFile(s), File object(s), and file(s) on disk are deleted. 

Also includes release note and tag bump for deployment, and two new tests for the new `delete_file_and_children()` function in `views_utils.py`. 

Screenshot:
image

[SYS-1598]: https://uclalibrary.atlassian.net/browse/SYS-1598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYS-1599]: https://uclalibrary.atlassian.net/browse/SYS-1599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oh_staff_ui/views_utils.py
+++ b/oh_staff_ui/views_utils.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from lxml import etree
 import requests
 import uuid
+from pathlib import Path
 from django.conf import settings
 from django.contrib import messages
 from django.core.management import call_command
@@ -35,6 +36,7 @@ from oh_staff_ui.models import (
     Date,
     Description,
     Format,
+    MediaFile,
     Name,
     ProjectItem,
     Subject,
@@ -613,3 +615,36 @@ def add_oai_envelope_to_mods(ohmods: OralHistoryMods) -> etree.Element:
 
 def user_in_oh_staff_group(user: User) -> bool:
     return user.groups.filter(name="Oral History Staff").exists()
+
+
+def delete_file_and_children(media_file: MediaFile) -> None:
+    # if file has child files, delete them first
+    children = list(MediaFile.objects.filter(parent=media_file))
+    for child in children:
+        # first delete file from file system
+        if child.file:
+            delete_file_from_filesystem(child.file.name)
+        else:
+            logger.warning(f"File {child.file_name} does not exist on the file system.")
+        child.delete()
+    # then delete the file (and its file from file system)
+    if media_file.file:
+        delete_file_from_filesystem(media_file.file.name)
+    else:
+        logger.warning(
+            f"File {media_file.file_name} does not exist on the file system."
+        )
+    media_file.delete()
+
+
+def delete_file_from_filesystem(file_name: str) -> None:
+    file_path = Path(settings.MEDIA_ROOT).joinpath(file_name)
+    if not file_path.exists():
+        logger.info(f"File {file_name} does not exist on the file system.")
+    else:
+        try:
+            file_path.unlink()
+            logger.info(f"Deleted {file_name} from file system.")
+        except Exception as ex:
+            logger.error(f"Failed to delete {file_name} from file system: {ex}")
+    return
--- a/oh_staff_ui/views_utils.py
+++ b/oh_staff_ui/views_utils.py
@@ -635,7 +635,7 @@ def delete_file(media_file: MediaFile, user: User) -> None:
         media_file.file.delete()
     else:
         logger.warning(
-            f"File {media_file.file.name} does not exist on the file system."
+            f"File {file_name} does not exist on the file system; deleting media object anyhow."
         )
     logger.info(f"File {file_name} deleted by user {user}.")
     media_file.delete()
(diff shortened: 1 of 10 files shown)


oh_staff_ui/views_utils.py
```diff
@@ -608,3 +611,42 @@ def add_oai_envelope_to_mods(ohmods: OralHistoryMods) -> etree.Element:
...
+def user_in_oh_staff_group(user: User) -> bool:
+    return user.groups.filter(name="Oral History Staff").exists()
+
+
+def delete_file_and_children(media_file: MediaFile) -> None:
+    # if file has child files, delete them first
+    children = list(MediaFile.objects.filter(parent=media_file))
+    for child in children:
+        # first delete file from file system
+        if child.file:
```
akohler, line 625: This check doesn't do what you expect.  Django `FieldFile` objects (which is what `MediaFile.file` contains) only have metadata about the file.  They **should** point to real files... but nothing enforces that beyond the instant the `FieldFile` was created.

For the logging to be correct, you need to build an appropriate `Path` object and check its existence.  But as noted elsewhere, the actual deletion of these should be done via `MediaFile.file.delete()`.

In the Django shell:
```
from pathlib import Path
from django.conf import settings
from oh_staff_ui.models import MediaFile
mf = MediaFile.objects.last()

mf.id, mf.file
# Values will vary, but this is via a very recent local import of production data
# (25005, <FieldFile: oh_static/pdf/submasters/21198-zz002kqdx9-1-submaster.pdf>)

# Django confirms that `mf.file`... is not `None`, but that's all:
bool(mf.file)
# True
bool(mf.file is None)
False
if mf.file: print("exists... or does it?")
# exists... or does it?

# But the file system disagrees
p = Path(settings.MEDIA_ROOT + "/" + mf.file.name)
# PosixPath('secret_media/oh_static/pdf/submasters/21198-zz002kqdx9-1-submaster.pdf')
p.exists()
# False

# My local copy of files does not include any PDFs, including the one
# pointed to by mf.file (wildcard here to avoid possible sequence mismatches)
# $ ll secret_media/oh_static/pdf/submasters/21198-zz002kqdx9*
# ls: cannot access 'secret_media/oh_static/pdf/submasters/21198-zz002kqdx9*': No such file or directory

# File system is correct.  Django data is correct... but locally, not connected to a matching file system.
```

akohler (changes requested):
@ztucker4 Overall, this is close.  
* The "Delete" button appears only when my user is in the correct group - even superuser admins don't get the button, unless they are also explicitly in the group.
* The confirmation dialog looks great, and lists the information I expect.
* Derivatives can be deleted separately, as well as masters.
* Deleting a derivative first, then deleting the master, works as expected.
  * We might want to reconsider allowing explicit derivative deletion eventually, if we can fix the parent/child relationships for migrated data, but for now it's correct.

The implementation itself needs some work.  It does have the correct end result, but there is some redundant / unneeded code for the actual file deletion.  Please see individual comments for details.

If anything's unclear, or if I misunderstood the code, please let me know.
Thanks --Andy

ztucker4:
Hi @akohler, this is ready for a second look. I believe the file deletion logic is fixed now - `delete_file_from_filesystem()` has been replaced everywhere with the standard `file.delete()`, and file existence checks before deletion are now done with `Path` objects. Thanks for your explanations of the issues!

ztucker4:
Hi @akohler, the requested changes are done. Logs now contain an `INFO` message for each deleted file: `File {file_name} deleted by user {user}.` Small supporting changes (to pass the user to the deletion function) for this were made in `views.py` and `tests.py`. Repeated code in `delete_file_and_children()` has been pulled out into a new `delete_file()` function.

akohler (approved):

akohler:
@ztucker4 This all looks great now.  I had to fix a merge conflict due to the "audit files" work I slipped in yesterday, and I tweaked the "file not found" log message to clarify that the object was being deleted anyhow, so now deleting a master image and its children looks like this (after manually deleting a submaster image file):
```
WARNING 2024-05-22 15:11:13,873 oh_staff_ui.views_utils views_utils File oh_static/submasters/21198-zz002kqd1v-2-submaster.jpg does not exist on the file system; deleting media object anyhow.
INFO 2024-05-22 15:11:13,873 oh_staff_ui.views_utils views_utils File oh_static/submasters/21198-zz002kqd1v-2-submaster.jpg deleted by user akohler.
INFO 2024-05-22 15:11:13,877 oh_staff_ui.views_utils views_utils File oh_static/nails/21198-zz002kqd1v-2-thumbnail.jpg deleted by user akohler.
INFO 2024-05-22 15:11:13,881 oh_staff_ui.views_utils views_utils File oh_masters/masters/21198-zz002kqd1v-2-master.tiff deleted by user akohler.
```

All merged, thanks for all your work on this!

PR #127, opened 2024-05-20
SYS-1098: Add file audit management command

akohler:
Implements [SYS-1098](https://uclalibrary.atlassian.net/browse/SYS-1098).  Bumps version to `v1.1.4` for deployment.

This PR adds a management command, `audit_media_files`.  It compares information about media files and reports differences via the application log.  It takes one mandatory argument:
* `DB`: Check all `MediaFile` objects and report on any where the real file pointed to does not exist, or has a different size from what's in the database
* 'FILES': Check all real files in `MEDIA_ROOT` (and all subdirectories) and report on any where there is no corresponding `MediaFile` object
  * Also reports if there are multiple `MediaFile` objects matching the real file, which should not happen but... who knows?
* `ALL`: Runs both the `DB` and the `FILES` audits

Mismatches are reported via `logs/application.log` and look like this:
```
# From the DB audit
WARNING 2024-05-20 16:17:51,544 oh_staff_ui.management.commands.audit_media_files audit_media_files FILE MISSING:       secret_media/oh_masters/pdf/masters/21198-zz0008z797-1-master.pdf       NOT FOUND       101709244       None    631

# From the FILES audit
WARNING 2024-05-20 16:17:53,736 oh_staff_ui.management.commands.audit_media_files audit_media_files DB MISSING: NOT FOUND       oh_static/nails/21198-zz002k355g-1-thumbnail.jpg        None    8189    None
```

For review after collecting data:
```
# Select the relevant log messages, and output just fields 6 to the end of line (space-delimited)
$ grep audit_media_files logs/application.log | cut -d" " -f6- > audit_trial.txt

FILE MISSING:       secret_media/oh_masters/pdf/masters/21198-zz0008z797-1-master.pdf       NOT FOUND       101709244       None    631

DB MISSING: NOT FOUND       oh_static/nails/21198-zz002k355g-1-thumbnail.jpg        None    8189    None
```
Remaining data is 6 fields, tab-delimited:
1. Description
2. DB file path, from `MediaFile.file.name`
3. Disk file path
4. DB file size, from `MediaFile.file_size`
5. Disk file size
6. Item id, from `MediaFile.item_id`

#### Testing
I chose not to write any tests. This command does not modify data, and outputs everything to a log file, and I opted not to contort things to make it testable. 

I tested it locally in 2 ways:
1. Using master and derivative images downloaded & generated during the image orientation review and cleanup, with all other `MediaFile` entries deleted from local database.  I confirmed that with this limited set, no discrepancies were found.  Then I manually renamed one file on disk (so it differed from its `MediaFile` data), and the script reported this difference correctly.
2. I imported a recent full production database dump locally.  The script reported 24,447 differences... since the vast majority of real files are **not** on my local system, and the image derivatives I generated locally have different sequence numbers from production, almost nothing matched.  This is expected.

So... please review the logic, make sure the command runs for you, and that should be enough.

[SYS-1098]: https://uclalibrary.atlassian.net/browse/SYS-1098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v1.1.3
+  tag: v1.1.4
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/management/commands/audit_media_files.py
+++ b/oh_staff_ui/management/commands/audit_media_files.py
@@ -1,6 +1,6 @@
 import logging
 from pathlib import Path
-from django.core.management.base import BaseCommand, CommandParser
+from django.core.management.base import BaseCommand, CommandParser, CommandError
 from django.conf import settings
 from oh_staff_ui.models import MediaFile
 
@@ -15,7 +15,8 @@ class Command(BaseCommand):
         parser.add_argument(
             "mode",
             type=str,
-            help="Mode: one of DB|FILES|BOTH",
+            help="Mode: one of DB|FILES|ALL",
+            choices=["DB", "FILES", "ALL"],
         )
 
     def handle(self, *args, **options) -> None:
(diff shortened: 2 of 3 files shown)


ztucker4 (approved):
This looks good to me. The logic of the command makes sense and covers cases of incorrect data in both directions (DB <-> MediaFiles). Results look correct when running locally. I see lots of missing files due to my copy of production data. I manually added and removed a couple of extra files and these were detected correctly. Merging now!

PR #131, opened 2024-05-29
SYS-1600: allow deletion of item records

ztucker4:
Implements [SYS-1600](https://uclalibrary.atlassian.net/browse/SYS-1600)

Adds the ability to delete ProjectItems without dependencies for users in the OH Staff user group. Users in this group will see a "Delete item" link within the "Item Actions" area of `edit_item` pages. Clicking this link will bring up a "Delete Item" modal. If the item has child ProjectItems or associated MediaFiles, the modal informs the user of these dependencies and the user has no option to delete. If the item can be deleted, the user is given the choice to "Delete" or "Cancel", with "Cancel" as the default. After deletion, if the item had a parent, the user is redirected to the parent's `edit_item` page. If not, the user is redirected to `item_search`. 

The underlying deletion view (`delete_item`) enforces that the requesting user must be in the OH Staff group, and that the item actually has no dependencies. In any case, a message is logged (indicating if an item was not deleted due to dependencies, if an unauthorized user attempts to use the deletion view directly, or if an item is deleted successfully). 

Also includes minor refactoring of the unrelated `upload_file` template to use better function names to control confirmation modals, four new tests (111 total), release notes and tag bump for deployment.

[SYS-1600]: https://uclalibrary.atlassian.net/browse/SYS-1600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v1.1.7
+  tag: v1.1.8
   pullPolicy: Always
 
 nameOverride: ""
--- a/oh_staff_ui/templates/oh_staff_ui/edit_item.html
+++ b/oh_staff_ui/templates/oh_staff_ui/edit_item.html
@@ -293,7 +293,7 @@
     {% bootstrap_button button_type="submit" content="Save" %}
 </form>
 
-<div class = "confirm-delete-popup" id="confirm-delete-popup">
+<div class="confirm-delete-popup" id="confirm-delete-popup">
     <h3>Delete Item</h3>
     {% if children_blocking_delete or media_files_blocking_delete %}
     <p>This item has dependencies and cannot be deleted.</p>
--- a/oh_staff_ui/templates/oh_staff_ui/edit_item.html
+++ b/oh_staff_ui/templates/oh_staff_ui/edit_item.html
@@ -313,10 +313,10 @@
         {% endfor %}
     </ul>
     {% endif %}
-    <a href="#" class="btn btn-primary" id="cancel-item" onclick="hideItemConfirmDeletePopup()">OK</a>
+    <a href="#" class="btn btn-primary" id="cancel" onclick="hideItemConfirmDeletePopup()">OK</a>
     {% else %}
     <p>Are you sure you want to delete this item: {{item.title}}?</p>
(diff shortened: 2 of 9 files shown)


akohler (changes requested):
@ztucker4 This is excellent!  Tests all pass, code makes sense and is well-commented, including division of work into the main view and supporting methods and refactoring.  The UI looks great, the permissions allow & block as expected, log messages are clear.

The only reason I'm not merging this immediately: While experimenting, I think I found a problematic workflow which could be frustrating for users.

1. Set up a basic Series -> Interview -> Session hierarchy and add files to the Session.
2. Say "Oops" and decide to delete the Session.
3. Get blocked, since files are attached.  Great.
4. Delete the files. Great.
5. I still want to delete the Session (item), but can't since I'm on the file management page (/upload_file/item_id).  
6. I click the browser's "back" button to go back to the Session item record (/item/item_id)
7. I click "Delete item", and... get denied, saying the files I just deleted still exist.
8. Since I know, I hit F5 to refresh the page, then try deletion again, and am allowed.

See screenshots below.

I'm not sure of the best way to solve this.  The root cause is that the "Delete Item" link does not hit the database to get current child / file information, and I doubt that trying that via javascript is the right approach.

I don't like telling users not to use the back button, or that they must refresh sometimes.

We could add a "Delete Item" link to the file management page... but that would mean passing unneeded data around, to somewhere it's not appropriate / needed most of the time.

What do you think would be the best way to approach this, ideally without requiring an awkward workflow?

Screenshot for steps 5-8 of my process above:
oh_step_5_after_deleting_files
oh_step_6_after_back_button
oh_step_7_denied_by_nonexistent_files
oh_step_8_after_manually_refreshing


Thanks --Andy

ztucker4:
I believe I've found a simple workaround to this issue: Django's `@never_cache` view decorator ([documentation here](https://docs.djangoproject.com/en/4.2/topics/http/decorators/#caching)). This forces the browser to reload the `edit_item` page even when accessed via the back button. Adding this decorator also helps prevent a confusing opposite situation to the one you found, in which a user might add a file to an item, use the back button, try to delete the item, and not be warned about dependencies.

akohler (approved):

akohler:
@ztucker4 Excellent solution with `@never_cache`.  Tested both scenarios, works as expected.  Merged!

PR #132, opened 2024-06-05
SYS-1609: Improve logging for file creation failures

akohler:
Implements [SYS-1609](https://uclalibrary.atlassian.net/browse/SYS-1609).  Bumps version to `v1.1.9` for deployment.

This PR improves handling of `MediaFile` creation, which all happens via `OralHistoryFile.process_media_files()`.  `MediaFile` creation now is wrapped in a `try` block which catches generic `OSError`s. This should handle errors with permissions, non-existent directories which can't be created by Django, and full disks.  These are now caught, logged, stored in the database as `MediaFileError`s to display in the UI, and then re-raised as `ValueError` to be handled gracefully by the UI.

It also fixes a small bug in `AudioFileHandler` where what should have been a multi-line string was actually two strings, so the second one was ignored.

Finally, it changes `MEDIA_ROOT` in the developer environment to `secret_media`, which is inside the volume mounted by Docker, instead of `/tmp/media_dev` which is not.  This makes it easy to interact with uploaded files from the developer's host system, instead of having to `docker cp` or shell into the container.

### Testing

I tried for hours to write a working test which covered this, but ultimately gave up.  I simulated `MEDIA_ROOT` being a directory owned only by `root`, so Django didn't have permissions; `MEDIA_ROOT` being full, by pointing it to `/dev/full` - but that's a file, not a directory, so didn't work as expected; and removing write permissions from the target directory within `MEDIA_ROOT`.  All failed... a `PermissionError` would get raised during the test, but could not be caught/handled via `assertRaises`, and then the failure would cascade to other tests.

However, testing manually works.  I manually changed permissions on a target directory, removing write permissions:
`chmod u-w secret_media/oh_masters/text/masters`

Then upload a text (XML) file to an item.  You should get a display like this:
oh_upload_failure

Then change permissions back to normal, and file upload should work now:
`chmod u+w secret_media/oh_masters/text/masters`

Since this class of error is unlikely to really happen (permissions and directories are all controlled in the container, and a full disk is unlikely though possible), I'm not spending more time trying to get a passing test.

[SYS-1609]: https://uclalibrary.atlassian.net/browse/SYS-1609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.docker-compose_django.env
+++ b/.docker-compose_django.env
@@ -35,7 +35,7 @@ DJANGO_OH_FILE_SOURCE=samples
 #DJANGO_OH_FILE_SOURCE=/home/django/oral-history-staff-ui/samples
 
 # Where Django looks for uploaded files
-DJANGO_OH_MEDIA_ROOT=/tmp/media_dev
+DJANGO_OH_MEDIA_ROOT=secret_media
 # Relative paths within oh_media_root, used by Django
 # Master files
 DJANGO_OH_MASTERS=oh_masters
--- a/charts/prod-ohstaff-values.yaml
+++ b/charts/prod-ohstaff-values.yaml
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: uclalibrary/oral-history-staff-ui
-  tag: v1.1.8
+  tag: v1.1.9
   pullPolicy: Always
 
 nameOverride: ""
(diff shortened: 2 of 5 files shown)


ztucker4 (approved):
This looks good to me. Manual testing works as expected - I can now see permissions errors that previously were unreported to the user. Comments are clear and helpful, and the change to `MEDIA_ROOT` is nice too. Merging now!

PR #135, opened 2024-06-17
SYS-1632: improve searching options

ztucker4:
Implements [SYS-1632](https://uclalibrary.atlassian.net/browse/SYS-1632) (mostly)

This PR adds two new functionalities to item search, which I believe satisfy [Lisa's requests](https://uclalibrary.atlassian.net/wiki/spaces/DEV/pages/276529172/OH+Staff+UI+Phase+3):
* Search filters for Item Type and Media File type. These can be used in conjunction with any search type to return only items with the specified Item Type or with attached Media Files of the specified type.
* Wildcard search. Searching for `*` will now return all items matching the filters. There is some simple JS validation to prevent users from performing a wildcard search without any filters (though the URL path `search_results/title/all/all/*` still exists and works).

I also have a few questions:
* This PR does not support one feature requested in the ticket: multiple selection for the item type filter. If this is really needed, I can implement this in the current search interface. It will increase the complexity of the URL parsing somewhat. 
* Should these searches be implemented using query strings instead of explicit URL paths? I feel like I usually see search URLs that look more like `search_results/title?itemtype=PDF&filetype=Interview&query=example` than my implementation of `search_results/title/PDF/Interview/example`. I'm not sure what the advantages of either approach are though.
* In Phase 2, we went to some lengths to make Status search mutually exclusive with Title, Keyword, or ARK search. I don't see why this is needed, and I think Status could easily be treated as another search filter alongside Item Type and Media File Tye. Do you think this should be added to this PR (or done in a separate ticket)?

[SYS-1632]: https://uclalibrary.atlassian.net/browse/SYS-1632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oh_staff_ui/views.py
+++ b/oh_staff_ui/views.py
@@ -111,11 +111,11 @@ def item_search(request: HttpRequest) -> HttpResponse:
             item_type_filter = form.cleaned_data["item_type_filter"]
             media_file_type_filter = form.cleaned_data["media_file_type_filter"]
             status_filter = form.cleaned_data["status_filter"]
-            if form.cleaned_data["item_type_filter"] is None:
+            if item_type_filter is None:
                 item_type_filter = "all"
-            if form.cleaned_data["media_file_type_filter"] is None:
+            if media_file_type_filter is None:
                 media_file_type_filter = "all"
-            if form.cleaned_data["status_filter"] is None:
+            if status_filter is None:
                 status_filter = "all"
 
             return redirect(
--- a/oh_staff_ui/views.py
+++ b/oh_staff_ui/views.py
@@ -106,8 +106,6 @@ def item_search(request: HttpRequest) -> HttpResponse:
                 typed_query = form.cleaned_data["status_query"]
             else:
                 typed_query = form.cleaned_data["char_query"]
-                # remove slashes from query to prevent url errors
-                typed_query = typed_query.replace("/", " ")
 
             # check for filter values, and set to known defaults if not present
             item_type_filter = form.cleaned_data["item_type_filter"]
--- a/oh_staff_ui/views.py
+++ b/oh_staff_ui/views.py
@@ -106,9 +106,22 @@ def item_search(request: HttpRequest) -> HttpResponse:
                 typed_query = form.cleaned_data["status_query"]
             else:
                 typed_query = form.cleaned_data["char_query"]
+                # remove slashes from query to prevent url errors
+                typed_query = typed_query.replace("/", " ")
+
+            # check for filter values, and set to known defaults if not present
+            item_type_filter = form.cleaned_data["item_type_filter"]
+            media_file_type_filter = form.cleaned_data["media_file_type_filter"]
+            if form.cleaned_data["item_type_filter"] is None:
+                item_type_filter = "all"
+            if form.cleaned_data["media_file_type_filter"] is None:
+                media_file_type_filter = "all"
+
             return redirect(
                 "search_results",
                 search_type=form.cleaned_data["search_type"],
+                item_type_filter=item_type_filter,
+                media_file_type_filter=media_file_type_filter,
                 query=typed_query,
             )
     else:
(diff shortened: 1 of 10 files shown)


oh_staff_ui/views_utils.py
```diff
@@ -117,30 +117,81 @@ def get_result_items(queryset_list: list) -> list[ProjectItem]:
...
-        results = [item for item in qs_results]
+    if isinstance(filtered_results, QuerySet):
+        results = [item for item in filtered_results]
     else:
-        results = qs_results
+        results = filtered_results
     return results
 
 
+def filter_keyword_results_list(
```
akohler, line 167: `filter_keyword_results_list()` and `filter_non_keyword_results_queryset()` are very similar, just acting on either a list or a queryset.  Can these be combined in a reasonable way, maybe by converting queryset to list, since that's what happens at the end of `get_search_results()` anyhow?

akohler (changes requested):
@ztucker4 This is great.  The only code change I'm requesting is to try to combine the two new filtering methods, since they have the same overall result - see my comment for details.

Otherwise, for the questions you raised and from our discussion:
1. Multiple selection is not needed, that was my generalization in the ticket.  Single selection on all filters is fine for now, we'll let the users ask for enhancements if they want to.
2. Let's stick with URL paths as implemented for now, unless you find that moving Status to a filter just makes that too unwieldy.  I'm fine with switching to URL parameters instead, so if you feel that's warranted, go ahead.
3. I agree that moving Status to an optional filter, working just like the new filters, is a good change.
4. Please add "Your search found `n` records" to the search result template.

Thanks --Andy

ztucker4:
Hi @akohler, this is ready for a second look. I've made the requested changes:
* Status is now a filter, like Media File Type and Item Type.
* The search results page now includes "Your search returned `N` results." above the results table.
* Search filtering code in `views_utils.py` is now simplified. As you suggested, there is now a single function (`filter_results_list()`) to perform filtering. If needed, querysets are converted to lists before this function is used.

oh_staff_ui/views.py
```diff
@@ -106,9 +106,24 @@ def item_search(request: HttpRequest) -> HttpResponse:
                 typed_query = form.cleaned_data["status_query"]
             else:
                 typed_query = form.cleaned_data["char_query"]
+
+            # check for filter values, and set to known defaults if not present
+            item_type_filter = form.cleaned_data["item_type_filter"]
```
akohler, line 111: These assignments & comparisons can be simplified.  Instead of reading `form.cleaned_data["some_filter"]` twice, this looks cleaner:
```
item_type_filter = form.cleaned_data["item_type_filter"]
if item_type_filter is None:
    item_type_filter = "all"
```
Could also be done with [ternaries](https://docs.python.org/3.10/reference/expressions.html#conditional-expressions) but I think that would be less readable here.

oh_staff_ui/views_utils.py
```diff
@@ -117,28 +117,73 @@ def get_result_items(queryset_list: list) -> list[ProjectItem]:
...
     # different return types and sort orders; this unifies them for consistent
     # use in the view and template.
-    if search_type == "title":
+
+    # first, check if this is a wildcard search - if so, return all items
+    # no need to check search_type
+    if query == "*":
+        qs_results = ProjectItem.objects.all().order_by("title")
+        results_list = [item for item in qs_results]
+        filtered_results = filter_results_list(
```
akohler, line 136: No need for this identical line in all 4 cases.  Since it does the same thing for each case, just move it below them, right above the `return` statement.

akohler (changes requested):
@ztucker4 The revised functionality works well.  A couple more minor code simplifications and this will be ready to merge.  Thanks --Andy

ztucker4:
Final changes made in `views` and `views_utils`, ready for one more review. Thanks!

akohler:
@ztucker4 Great, thanks - merged.

PR #25, opened 2023-03-24
SYS-1192: Update models for uncontrolled metadata

akohler:
Implements [SYS-1192](https://jira.library.ucla.edu/browse/SYS-1192).

This PR changes several models related to uncontrolled metadata - data which may have qualifying types, but the values are not controlled and may be freely edited.

For these 3, existing models were updated to fix incorrect initial assumptions.  Source and usage were removed, leaving types (lookup tables) and main tables which associated values with items as standard one-to-many foreign key relations:
* `AltID`
* `AltTitle`
* `Description`

A new set of models was added for Dates, also with lookup types and standard 1-many FK  item relations:
* `Date`
* `DateType`

Finally, the `ProjectItem` model was updated to add an optional `relation` field.

Related changes:
* New migration file `0015...`: supports the above model changes.
* `admin.py`
  * removed displays for uncontrolled data tables, including `Format`.  Editors should interact with these only through the `edit_item` form.
  * removed display of `ProjectItem`.  This also is uncontrolled data which goes through `edit_item`, and has minimal value for developers in the admin interface.
  * added display of `DateType` data.
* Deleted fixtures which provided minimal sample data for these models.  The association with items was increasingly hard to maintain manually, and could conflict with real data being loaded (for developers) soon.  The corresponding qualifier type fixtures are sill loaded.
* Added fixture for `DateType` data.
* `entrypoint.sh`: updated to match fixture changes, and slight reorganization of remaining fixture loads.

### Testing
Mainly, make sure model changes are correct, so changes to the editing form and the data loading scripts are completely unblocked.

Since this involves migration and fixture loading changes, restarting docker-compose system is best:
```
docker-compose down
docker-compose up -d
# check logs... should be no errors other than the standard "superuser exists"
docker-compose logs -f django
```

If any errors, delete database and try again:
```
docker-compose down
docker volume rm oral-history-staff-ui_pg_data
docker-compose up -d
```

PR #144, opened 2026-01-30
SYS-2156: Allow item deletion without forcing manual deletion of descriptions

akohler:
Fixes [SYS-2156](https://uclalibrary.atlassian.net/browse/SYS-2156). Tagged as version `1.1.19` for release.

This PR fixes a bug which prevented `ProjectItem` records from being deleted if any `Description` records were linked to them.  The foreign key relation between `ProjectItem` and `Description` was set to `PROTECT`, which prevented the parent item from being deleted if it had any child descriptions.

This behavior appears to be left over from #25 , when `Description` and some other models changed to be purely 1-to-many, uncontrolled vocabulary.  Initially they were (potentially) many-to-many controlled vocabulary, so a `Description` could be associated with multiple items, and thus were protected from automatic deletion.  

Once I understood the history and correct normal usage, I changed the FK behavior from `PROTECT` to `CASCADE` (for `Description` only).  Now if an item has description(s), the item can be deleted without requiring the user to delete these dependent records.  This is a trivial change in the model, and a migration to implement it.

I also added 2 tests to the existing `ItemDeletionTestCase`: 1 to confirm that items with descriptions can be deleted without raising an error (thus no `assert` in `test_projectitem_with_description_is_deletable()`), and 1 to confirm that items with `PROTECT`ed relations like `Name` *cannot* be deleted as-is.

All the other changes in the PR are incidental:
* Added our standard Django-specific devcontainer configuration
* Fixed some unrelated type errors in HTML tests which used `str` instead of the expected `SafeText` type
* Fixed some unrelated return types in XML utility functions which were showing type errors in their associated tests.

I chose *not* to apply the `PROTECT` -> `CASCADE` change to a couple of other models (`AltID` and `AltTitle`) where it probably would be appropriate, because these are used far less often and have not been reported as a problem... so I punted to future-somebody, possibly.

I also chose not to go down the rabbit-hole of fixing every type error still showing in `views_utils.py`, mostly related to XML and the lack of built-in type hinting in `lxml` and `eulxml` - it just isn't worth it, until / unless we actually have to touch that code.  I fixed only the minimum to make `pylance` happy with `tests.py`.

### Testing

There now are 114 tests; all should pass.  No manual testing is necessary, though you're welcome to set up a test item with description and possibly a controlled term like Name or Subject, and confirm the manual behavior is the same as the tests.

Note that running tests shows this warning:
```
/home/django/.local/lib/python3.13/site-packages/eulxml/__init__.py:19: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```
This is due to code in the probably no longer maintained `eulxml`.  We already pin `setuptools` to an older value in `requirements.txt`.  I tried the more generic `< 81` just in case; it did not remove this warning.  We could probably suppress the warning but I'd rather it show up so we're more aware if / when something really does break in the future.

[SYS-2156]: https://uclalibrary.atlassian.net/browse/SYS-2156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ztucker4 (approved):
This all looks good to me! Changing the relevant `on_delete` to `CASCADE` makes sense and works as expected.

I manually recreated the non-working scenario, and then confirmed that the I can delete ProjectItems with Descriptions using the new code. The automated tests also make look good and pass.

Also, thanks for setting up the devcontainer config and fixing type errors (TIL about `SafeText`). Merging now!

PR #145, opened 2026-05-07
[DLSR-325] Security patches, README cleanup

DRickard:
* Update `django`, `pillow`, `request` versions for security patching
* Update README to reflect new Django version, replace outdated `docker-compose` with `docker compose`
* Update Helm chart/release notes to reflect security patches
* Output from `docker-compose exec django python manage.py test` execution:
```
/home/django/.local/lib/python3.13/site-packages/eulxml/__init__.py:19: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
Found 114 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
..................................................................................................................
----------------------------------------------------------------------
Ran 114 tests in 3.387s

OK
Destroying test database for alias 'default'...

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Related repositories:
 The development environment requires:
 * git (at least version 2)
 * docker (current version recommended: 20.10.12)
-* docker-compose (at least version 1.25.0; current recommended: 1.29.2)
+* docker compose (at least version 1.25.0; current recommended: 1.29.2)
 
 #### PostgreSQL container
 
@@ -30,7 +30,7 @@ The development database is a Docker container running PostgreSQL 12, which matc
 
 #### Django container
 
-This uses Django 4.1, in a Debian 11 (Bullseye) container running Python 3.11.  All code 
+This uses Django 5.2.13, in a Debian 11 (Bullseye) container running Python 3.14.  All code 
 runs in the container, so local version of Python does not matter.
 
 The container runs via `docker_scripts/entrypoint.sh`, which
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,9 +1,9 @@
-Django == 5.2.10
+Django >= 5.2.13
```
akohler: @DRickard For versions in `requirements.txt`, we almost always pin specific versions - the `==` as opposed to `>=`.  That way, the application changes only when we deliberately update these, instead of potentially changing from build to build.

README.md
```diff
@@ -22,15 +22,15 @@ Related repositories:
 The development environment requires:
 * git (at least version 2)
 * docker (current version recommended: 20.10.12)
-* docker-compose (at least version 1.25.0; current recommended: 1.29.2)
+* docker compose (at least version 1.25.0; current recommended: 1.29.2)
 
 #### PostgreSQL container
 
 The development database is a Docker container running PostgreSQL 12, which matches our deployment environment.
```
akohler: Per `docker-compose.yml`, we're on postgres 16:
```
  db:
    image: postgres:16
```

README.md
```diff
@@ -22,15 +22,15 @@ Related repositories:
...
 
 #### Django container
 
-This uses Django 4.1, in a Debian 11 (Bullseye) container running Python 3.11.  All code 
+This uses Django 5.2.13, in a Debian 11 (Bullseye) container running Python 3.14.  All code
```
akohler: Per `Dockerfile`, we're on Debian 13 (Bookworm) running Python 3.13:
```
FROM python:3.13-slim-bookworm
```

akohler (changes requested):
@DRickard This is good.  I confirmed the updated images builds and all tests pass.

The `README` needs a few updates still, to the correct OS/Python versions and an update to the Postgres version which I didn't mention in the ticket.

Also, please pin specific versions of packages in `requirements.txt`, to avoid unexpected changes.  We could consider specifying maximum versions as well, to allow some flexibility, but we currently don't do that.  With unbounded versions like `Django >= 5.2.13`, the next person building this could wind up with Django 6.

Once those minor corrections are made, this will be ready to merge.  Thanks --Andy

akohler (approved):
@DRickard All good now, merging & deploying, thanks.

